PR #2, opened 2024-04-26
demo: add share button

truckermudgeon:
@nautofon [mentioned](https://gist.github.com/nautofon/d6b3fb841f632478c6db6f0d7f00231e#ui) the useful idea of:

> A marker you can drop, with a link you can share, so that others can see the exact place you're referring to. Something like [markers on osm.org](https://www.openstreetmap.org/?mlat=32.7772&mlon=-99.8973#map=11/32.7772/-99.9485).

This PR does that by adding a Share button to the set of top-right map controls:

https://github.com/truckermudgeon/maps/assets/121829201/215cb343-330c-4d61-9206-c90e61d44f80

A few issues that may be addressed in the future:
  * shared links don't preserve the current set of Map Options
  * the only way to clear a Marker after visiting a page with a Marker is to manually edit the URL in the browser
  * the `routes` demo doesn't have a Share button

### DIFF
--- a/packages/apps/demo/src/Demo.tsx
+++ b/packages/apps/demo/src/Demo.tsx
@@ -22,14 +22,16 @@ import { MapSelectAndSearch } from './MapSelectAndSearch';
 import { ShareControl } from './ShareControl';
 import { toStateCodes } from './state-codes';
 
+const inRange = (n: number, [min, max]: [number, number]) =>
+  !isNaN(n) && min <= n && n <= max;
+
 const Demo = () => {
   const [searchParams] = useSearchParams();
+  const lat = Number(searchParams.get('mlat'));
+  const lon = Number(searchParams.get('mlon'));
   const markerPos =
-    searchParams.has('mlat') && searchParams.has('mlon')
-      ? {
-          lat: Number(searchParams.get('mlat')),
-          lon: Number(searchParams.get('mlon')),
-        }
+    inRange(lat, [-90, 90]) && inRange(lon, [-180, 180])
+      ? { lat, lon }
       : undefined;
 
   const [autoHide, setAutoHide] = useState(true);
--- a/packages/apps/demo/src/Demo.tsx
+++ b/packages/apps/demo/src/Demo.tsx
@@ -12,13 +12,26 @@ import { useState } from 'react';
 import MapGl, {
   AttributionControl,
   FullscreenControl,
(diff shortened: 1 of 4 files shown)


packages/apps/demo/src/Demo.tsx
```diff
@@ -12,13 +12,26 @@ import { useState } from 'react';
...
+import { ShareControl } from './ShareControl';
 import { toStateCodes } from './state-codes';
 
 const Demo = () => {
+  const [searchParams] = useSearchParams();
+  const markerPos =
+    searchParams.has('mlat') && searchParams.has('mlon')
+      ? {
+          lat: Number(searchParams.get('mlat')),
+          lon: Number(searchParams.get('mlon')),
```
truckermudgeon: Oof... looks like I need to do some validation/cleanup. When manually entering a URL with junk params:

image

packages/apps/demo/src/Legend.tsx
```diff
@@ -103,21 +103,20 @@ export const Legend = (props: LegendProps) => {
   const [activeTab, setActiveTab] = useState<number>(0);
   return (
     <>
-      <Tooltip title={'Map Options'}>
```
truckermudgeon, line 106: Unrelated change:

This removes the `<Tooltip>` wrapper and adds a `title` attribute to the IconButton so that the Map Options button looks/feels like the other MapLibre buttons.

packages/apps/demo/src/Legend.tsx
```diff
@@ -131,7 +130,6 @@ export const Legend = (props: LegendProps) => {
           },
           content: {
             sx: {
-              pointerEvents: 'auto',
```
truckermudgeon, line 134: Another unrelated-to-sharing change:

Moved restoration of pointer events deeper down in the tree. Turns out pointer events were being intercepted by the Drawer in the areas above/below the drawer content.

packages/apps/demo/src/Demo.tsx
truckermudgeon: https://github.com/truckermudgeon/maps/pull/2/commits/43046809d61e8c9fcdf17654530ad0c6218bdabd

johannessen:
Oh! Very cool. 😎

nautofon:
Oh! Very cool. 😎

PR #3, opened 2024-05-03
parser: switch over to HashFS v2

truckermudgeon:
This PR updates `parser` to support [HashFS v2](https://modding.scssoft.com/wiki/Documentation/Tools/Game_Archive_Packer) .scs archives. Much of the parsing code is adapted from the [ConverterPIX](https://github.com/mwl4/ConverterPIX/) project. 

The trickiest bit with HashFS v2 is how .tobj/.dds file pairs are treated: in HashFS v1, they're separate file entries in an .scs archive. In HashFS v2, a single entry is present, and it only contains the .dds data bytes. `parser` does some extra work to reconstruct the .dds header so that texture data (for the various map icons) can be read.

Here's a screenshot of some of the California Rework changes in the 1.50 experimental beta, parsed with the code in this PR:

image

### DIFF
--- a/packages/clis/parser/game-files/scs-archive.ts
+++ b/packages/clis/parser/game-files/scs-archive.ts
@@ -203,7 +203,7 @@ export class ScsArchive {
   ): Map<number, MetadataEntry> {
     const metadataMap = new Map<number, MetadataEntry>();
 
-    const metadataBuffer = this.readData({
+    const metadataTable = this.readData({
       offset: this.header.metadataTableOffset,
       compressedSize: this.header.metadataTableCompressedSize,
       uncompressedSize: this.header.metadataTableSize,
@@ -211,37 +211,45 @@ export class ScsArchive {
     const skippedMetaTypes = new Set();
     for (const header of entryHeaders) {
       for (let i = 0; i < header.metadataCount; i++) {
-        const startOffset = 4 * (header.metadataIndex + i);
-        const slice = metadataBuffer.subarray(
-          startOffset,
-          startOffset + MetadataEntryHeader.size(),
+        const metadataHeaderByteOffset = 4 * (header.metadataIndex + i);
+        const metadataHeader = MetadataEntryHeader.fromBuffer(
+          metadataTable.subarray(
+            metadataHeaderByteOffset,
+            metadataHeaderByteOffset + MetadataEntryHeader.size(),
+          ),
         );
-        const h = MetadataEntryHeader.fromBuffer(slice);
-        const type = h.type as MetadataType;
+        const type = metadataHeader.type as MetadataType;
         switch (type) {
           case MetadataType.IMG:
           case MetadataType.SAMPLE:
           case MetadataType.PLAIN:
           case MetadataType.DIRECTORY:
           case MetadataType.MIP_TAIL: {
-            const descriptor =
-              type === MetadataType.IMG
-                ? ImageMeta
-                : type === MetadataType.SAMPLE
-                  ? SampleMeta
-                  : PlainMeta;
-            const meta = {
-              ...descriptor.fromBuffer(metadataBuffer.subarray(4 * h.index)),
-              version: h.type,
-            };
-            metadataMap.set(header.metadataIndex + i, meta);
+            let descriptor;
+            if (type === MetadataType.IMG) {
+              descriptor = ImageMeta;
+            } else if (type === MetadataType.SAMPLE) {
+              descriptor = SampleMeta;
+            } else {
+              descriptor = PlainMeta;
+            }
+            const metadataEntryByteOffset = 4 * metadataHeader.index;
+            metadataMap.set(header.metadataIndex + i, {
+              version: metadataHeader.type,
+              ...descriptor.fromBuffer(
+                metadataTable.subarray(
+                  metadataEntryByteOffset,
+                  metadataEntryByteOffset + descriptor.size(),
+                ),
+              ),
+            });
             break;
           }
           case MetadataType.MIP_0:
           case MetadataType.MIP_1:
           case MetadataType.MIP_PROXY:
           case MetadataType.INLINE_DIRECTORY:
-            skippedMetaTypes.add(h.type);
+            skippedMetaTypes.add(metadataHeader.type);
             break;
         }
       }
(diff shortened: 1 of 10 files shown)


packages/clis/parser/gdeflate/gdeflate.cc
```diff
@@ -0,0 +1,49 @@
...
+  Napi::Env env = info.Env();
+
+  if (info.Length() != 2) {
+    Napi::TypeError::New(env, "Wrong number of arguments")
+        .ThrowAsJavaScriptException();
+    return env.Null();
+  }
+  if (!info[0].IsArrayBuffer() || !info[1].IsArrayBuffer()) {
+    Napi::TypeError::New(env, "Arguments must be array buffers").ThrowAsJavaScriptException();
+    return Napi::Value::From(env, -1);
```
truckermudgeon: If we're throwing a JS exception above, then the return value probably doesn't matter.

```suggestion
    return env.Null();
```

packages/clis/parser/types/restructure.d.ts
```diff
@@ -99,11 +99,13 @@ declare module 'restructure' {
   }
 
   export type StructType<T extends Record<string, unknown>> = {
-    [K in keyof T]: T[K] extends Base<infer U>
-      ? U
-      : T[K] extends () => infer V
-        ? V
-        : T[K];
+    [K in keyof T]: T[K] extends Base<never>
+      ? undefined
```
truckermudgeon, line 103: Added this so that struct literals with reserved fields just need to specify `undefined` for a value.

E.g., `DdsHeader` has a `reserved1: new r.Reserved(r.uint32le, 11)` field. Before this change, calling `DdsHeader::toBuffer` required a struct literal with a `reserved1: [0,0,0,0,0,0,0,0,0,0,0]` field in order for type checks to pass. Now, the struct literal can just have `reserved1: undefined`.

Ideally, it'd be possible to just omit the `reserved1` field altogether from the struct literal. But I can't be bothered to figure out how to do that 🙃.

packages/clis/parser/types/restructure.d.ts
```diff
@@ -53,7 +53,7 @@ declare module 'restructure' {
     size(): number;
   }
 
-  export class Enum<T extends readonly string[]> extends Base<
+  export class Enum<T extends readonly unknown[]> extends Base<
```
truckermudgeon: Gonna revert this change... never needed to make use of it (and `unknown` is too broad a type).
```suggestion
  export class Enum<T extends readonly string[]> extends Base<
```

packages/clis/parser/game-files/scs-archive.ts
```diff
@@ -1,34 +1,132 @@
...
+};
+
 const FileHeader = new r.Struct({
   magic: new r.String(4),
   version: r.int16le,
   salt: r.int16le,
   hashMethod: new r.String(4),
-  numEntries: r.int32le,
-  entriesOffset: r.int32le,
+  entryTableCount: r.uint32le,
```
truckermudgeon, line 28: 🤔 I should probably restore HashFS v1 support (in addition to adding HashFS v2 support). 

It might take a few months for 1.50 to officially launch, and I'd rather have this code merged before then. Also: it's probably friendlier to have a "latest 2 versions" support policy instead of a "latest version, only" policy.

packages/clis/parser/game-files/scs-archive.ts
```diff
@@ -104,19 +198,77 @@ export class ScsArchive {
     return this.entries;
   }
 
+  private createMetadataMap(
+    entryHeaders: BaseOf<typeof EntryHeader>[],
+  ): Map<number, MetadataEntry> {
+    const metadataMap = new Map<number, MetadataEntry>();
+
+    const metadataBuffer = this.readData({
```
truckermudgeon: Clearer if this were a Uint32Array?

packages/clis/parser/game-files/scs-archive.ts
```diff
@@ -104,19 +198,77 @@ export class ScsArchive {
...
+          case MetadataType.SAMPLE:
+          case MetadataType.PLAIN:
+          case MetadataType.DIRECTORY:
+          case MetadataType.MIP_TAIL: {
+            const descriptor =
+              type === MetadataType.IMG
+                ? ImageMeta
+                : type === MetadataType.SAMPLE
+                  ? SampleMeta
+                  : PlainMeta;
```
truckermudgeon: Maybe not worth doing a nested ternary 😬

packages/clis/parser/game-files/scs-archive.ts
```diff
@@ -130,20 +282,86 @@ function createStore<V extends { hash: bigint }>(values: V[]) {
...
+
   return metadata.isDirectory
     ? new ScsArchiveDirectory(fd, metadata)
     : new ScsArchiveFile(fd, metadata);
 }
 
+function createTobjEntry(
+  fd: number,
+  header: BaseOf<typeof EntryHeader>,
+  metadataMap: Map<number, MetadataEntry>,
```
truckermudgeon, line 339: How would things look if this were a sparse array instead of a Map?

truckermudgeon:
Note to self: update README file with:
* instructions that mention init'ing git submodule
* credit to https://github.com/mwl4/ConverterPIX

Edit: done in https://github.com/truckermudgeon/maps/pull/3/commits/79d3b2abacd64f55fc1fb6b7fb9cc531e5c628e9

packages/clis/parser/game-files/scs-archive.ts
nautofon: The general sentiment over on the forum seems to be that the Nebraska release will happen this month. 1.50 is all but certain to have been launched before then.

I suspect that at this point, v1 support is a case of [YAGNI](https://en.wikipedia.org/wiki/YAGNI).

truckermudgeon: Ooooo... thanks for the heads-up. In that case, I'll continue on as I originally planned (with the much-easier to maintain "latest version, only" policy 😁)

packages/clis/parser/game-files/scs-archive.ts
truckermudgeon: The answer: nope. Gotta do extra work to ensure little-endian-ness 😬.

PR #10, opened 2024-07-10
generator: add contours command

truckermudgeon:
This PR adds a `contours` command:

```sh
# generate ATS and ETS2 contours (aka elevations) pmtiles files
npx generator contours -m usa -m europe -i dirWithParserOutput -o dirToWriteFilesTo
```

which can be read by the `Contours` component added in #7. All the hard work is done by the [d3-delaunay](https://github.com/d3/d3-delaunay) library.

----

### Notes:

The `contours` command performs some naive masking based on sector bounds:

Screenshot 2024-07-11 at 6 01 16 PM

Screenshot 2024-07-11 at 6 01 57 PM

which looks less weird than the unmasked data that was originally published in https://github.com/truckermudgeon/truckermudgeon.github.io/commit/191724174cfba759d0d3861aeffb2e1be304328f:

Screenshot 2024-07-11 at 6 12 10 PM

Screenshot 2024-07-11 at 6 10 58 PM

For the eagle-eyed readers who are wondering about elevation differences in the screenshots above: the originally published contour data was calculated in QGIS by projecting game coordinates into WGS84 first, then calculating contours with a plugin. The code in this PR calculates contours against game coordinates first, then projects those contours into WGS84.

### DIFF
--- a/packages/clis/generator/geo-json.ts
+++ b/packages/clis/generator/geo-json.ts
@@ -756,6 +756,7 @@ export function convertToContoursGeoJson({
 
   logger.start(
     'calculating',
+    map,
     levels,
     'contour levels',
     `(${min} min, ${max} max)`,
@@ -786,6 +787,7 @@ export function convertToContoursGeoJson({
     );
     bar.increment();
   }
+
   logger.success(
     levels,
     'contours calculated in',
--- a/packages/clis/generator/geo-json.ts
+++ b/packages/clis/generator/geo-json.ts
@@ -56,6 +56,7 @@ import { tricontour } from 'd3-tricontour';
 import fs from 'fs';
 import type { GeoJSON } from 'geojson';
 import path from 'path';
+import polygonclipping from 'polygon-clipping';
 import url from 'url';
 import { normalizeDlcGuards } from './dlc-guards';
 import { logger } from './logger';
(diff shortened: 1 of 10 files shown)


packages/clis/generator/index.ts
```diff
@@ -374,6 +423,95 @@ function handleFootprintsCommand(
...
+    let cleanupGeoJson = false;
+    if (geoJsonPath == null) {
+      //geoJsonPath = path.join(os.tmpdir(), filename);
+      geoJsonPath = path.join(args.outputDir, filename);
+      logger.log('writing temporary GeoJSON file...');
+      writeGeojsonFile(geoJsonPath, geoJson);
+      cleanupGeoJson = true;
+    }
+
+    const minAttributes = ['elevation'];
```
truckermudgeon, line 462: Reminder to update [`Contours`](https://github.com/truckermudgeon/maps/pull/7/files#diff-dcc4891cb441c22f76dc3402c34e9173c0cc8913cc0d6f84999cba4c86f042ccR16) component to reference `elevation` (absolute values) instead of `index` (relative values)

packages/clis/generator/geo-json.ts
```diff
@@ -731,6 +735,105 @@ export function convertToFootprintsGeoJson({
...
+    if (p[2] < min) {
+      min = p[2];
+    }
+    if (p[2] > max) {
+      max = p[2];
+    }
+  }
+  const levels = max - min + 1;
+
+  logger.log('calculating sector mask...');
```
truckermudgeon, line 758: I _really_ wanted to prevent interpolating elevation data outside of sector bounds by making use of [constrained delaunay triangulation](https://en.wikipedia.org/wiki/Constrained_Delaunay_triangulation): `d3-tricontours` lets you specify a custom triangulation function, and it would've been great if I could just plug-in [poly2tri.js](https://github.com/r3mi/poly2tri.js).

But poly2tri.js appears to do its triangulation recursively and choked on the millions of points I gave it. And I didn't want to spend any more time on dealing with that issue, given how inaccurate the elevation data I'm using is, anyway (I'm gathering elevation data so naively and incompletely).

So simple polygon masking it is!

PR #11, opened 2024-07-16
generator: make checkGraph multi-threaded

truckermudgeon:
This PR reduces the time it takes to run the `generator graph` command with the `--check` option.

`--check` validates the generated graph by picking a random "origin" company and finding routes to/from all other companies. Any companies that cannot be routed to/from the origin company are logged for further examination.

Currently, route-checking is done one route at a time. This PR makes it so that:
* route checks are done with a thread pool
* unrouteable routes are reported at the end of the check
  * currently, they're reported as they're encountered, which messes with how the progress bar is rendered

The end result is a speedup in graph checking (on my machine: ~4x) and slightly nicer-looking console output. The speedup should especially be helpful as more routing-related work is done for the `navigator` project.


packages/clis/generator/graph/find-route-worker-wrapper.js
```diff
@@ -0,0 +1,32 @@
+import { workerData as tinypoolData } from 'node:worker_threads';
+import { register } from 'tsx/esm/api';
+
+// This file exists so that:
```
truckermudgeon, line 4: N.B.: this file may end up becoming a generic worker wrapper, if more `generator` commands end up being made multi-threaded (e.g., [contour generation](https://github.com/truckermudgeon/maps/blob/7149d69d1871aa52638610c248d6511eaf76836c/packages/clis/generator/geo-json.ts#L808-L822)).

packages/libs/map/routing.ts
```diff
@@ -108,21 +124,18 @@ export function findRoute(
...
-    'and',
-    goal,
-    `(${numIters} iterations)`,
-  );
-  return undefined;
+  return {
+    success: false,
+    key,
+    mode,
+    numIters,
```
truckermudgeon, line 131: The value of `numIters` is usually a good indicator of _where_ routing failed. If the value is high (like, 250k), then the graph's problem is likely near the destination (aka `end`) node. But if it's low (like, less than 100), the problem is likely near the origin (aka `start`) node.

As an example, one unrouteable origin/destination company pair currently being flagged is

```json
{
  "numIters": 258196,
  "start": {
    "company": "ali_car_dlr.hays",
    "nodeUid": "541ac1020c853cdb",
    "latLng": "/38.984/-99.311"
  },
  "end": {
    "company": "syn_che_plnt.st_joseph",
    "nodeUid": "541ac10150054bfe",
    "latLng": "/39.703/-95.177"
  }
}
```

The high `numIters` value suggests the problem lies with some company in St Joseph... which is in Missouri, which is still unreleased, and isn't surprising that no route to it could be found (especially since there are no roads connecting it to the rest of the road network!)

tsconfig.base.json
```diff
@@ -12,7 +12,8 @@
     //"noUnusedLocals": true,
     "noUnusedParameters": true,
     "isolatedModules": true,
-    "checkJs": true,
+    "allowJs": true,
```
truckermudgeon, line 15: Changes made to support the worker-wrapper JS file.

Had to change `checkJs` to `false`, otherwise `tsc` would throw errors that I didn't know how to fix 🫠:

```
packages/clis/generator/graph/find-route-worker-wrapper.js:2:26 - error TS2307: Cannot find module 'tsx/esm/api' or its corresponding type declarations.
  There are types at '/Users/icosahebron/dev/maps/node_modules/tsx/dist/esm/api/index.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.

2 import { register } from 'tsx/esm/api';
                           ~~~~~~~~~~~~~

packages/clis/generator/graph/find-route-worker-wrapper.js:30:12 - error TS7006: Parameter 'options' implicitly has an 'any' type.

30   default: options =>
              ~~~~~~~
```

For the first error: 🤷 . I still don't quite understand module resolution.

For the second error: how do you deal with such a type error in JS, where you can't add any type annotations 😳

PR #13, opened 2024-09-15
parser,generator: support 1.51 release

truckermudgeon:
This PR updates:
* `parser` to support version 901 map files
* `generator`'s ETS2 village CSV with krmarci's latest updates


packages/clis/parser/game-files/sector-parser.ts
```diff
@@ -258,6 +262,8 @@ const SimpleItemStruct = {
     width: r.floatle,
     height: r.floatle,
     fogMaskPresetId: r.int32le,
+    // new in v901.
+    unknown: new r.Reserved(r.uint8, 16),
```
truckermudgeon, line 266: I haven't looked to deeply to see if this new data should be treated as a dynamically-sized array... but parsing _seems_ to work when treating it as a static set of bytes 🙃

PR #15, opened 2024-10-03
Fix broken git clone link.

Tumppi066:
Using `git@github.com:truckermudgeon/maps.git` is not possible as it results in a permission error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ A demo app built with these tools and components can be found at https://trucker
 # clone the repo into a local directory
 mdkir maps
 cd maps
-git clone --recurse-submodules git@github.com:truckermudgeon/maps.git .
+git clone --recurse-submodules https://github.com/truckermudgeon/maps.git .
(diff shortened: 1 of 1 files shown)


truckermudgeon (approved):
Interesting... are you cloning on a machine without `ssh`? Or maybe you have `ssh` set up, but your key isn't password-protected?

Either way: LGTM, because HTTPS seems to be the easiest option.

nautofon:
FWIW, GitHub has a troubleshooting page for this particular error:

https://docs.github.com/en/authentication/troubleshooting-ssh/error-permission-denied-publickey

Tumppi066:
> FWIW, GitHub has a troubleshooting page for this particular error:
> 
> https://docs.github.com/en/authentication/troubleshooting-ssh/error-permission-denied-publickey

Well there we go that would've helped👍 
I think using https will be better overall as that will fix the issue too.

PR #17, opened 2024-10-04
Add shoulderSpace to road defs, add Z values to all items.

Tumppi066:
Feel free to ignore this PR if you have these features in a branch somewhere. This was done so that I can start moving over from ts-map before you add the features officially.

**I have not tested that the geojson conversion works still.** I am only adding more data, but I don't know how resiliant that code is to it.

### DIFF
--- a/packages/clis/parser/game-files/prefab-ppd-parser.ts
+++ b/packages/clis/parser/game-files/prefab-ppd-parser.ts
@@ -313,6 +313,7 @@ export function parsePrefabPpd(buffer: Buffer): PrefabDescription {
       return {
         x: n.pos[0],
         y: n.pos[2],
+        z: n.pos[1],
         rotation,
         inputLanes: n.inputLanes.filter(v => v !== -1),
         outputLanes: n.outputLanes.filter(v => v !== -1),
@@ -332,6 +333,7 @@ export function parsePrefabPpd(buffer: Buffer): PrefabDescription {
         const baseProperties = {
           x: pos[0],
           y: pos[2],
+          z: pos[1],
           neighbors: rp.neighbors.slice(0, neighborCount),
         };
 
@@ -411,8 +413,8 @@ export function parsePrefabPpd(buffer: Buffer): PrefabDescription {
         endPos,
         endRot,
       } = rc;
-      const start = getXYR(startPos, startRot);
-      const end = getXYR(endPos, endRot);
+      const start = getXYZR(startPos, startRot);
+      const end = getXYZR(endPos, endRot);
       return {
         navNodeIndex,
         start,
@@ -442,16 +444,16 @@ export function parsePrefabPpd(buffer: Buffer): PrefabDescription {
   };
 }
 
-function getXYR(
+function getXYZR(
   pos: [number, number, number],
   rot: [number, number, number, number],
 ) {
-  const [x, , y] = pos;
+  const [x, z, y] = pos;
   // TODO probably not the right way to do quat => euler.
   const [rx, , rz] = rot;
   // theta starts normally, but increases CW instead of CCW since game's coord system
   // has Y growing downwards.
   let rotation = Math.PI - Math.atan2(rz, rx);
   rotation = (rotation % Math.PI) * 2 - Math.PI / 2;
-  return { x, y, rotation };
+  return { x, y, z, rotation };
 }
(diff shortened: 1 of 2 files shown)


truckermudgeon:
There's a lot of lines here that look familiar 🙂 

I'll have my branch up + merged with the prefab z-values up Soon™... I added them while experimenting with rendering overlapping roadways:

image

I've also got another branch with the `shoulderSpace` values... somewhere.

As for adding z-values to map items: I've got two questions.

### How will z-values be used for Cities, Countries, and Ferries 🤔?

My understanding:

* Country position determines where a country's label appears in the world map. In ATS, all elevation values are `0`:
```
country % grep pos -R
./utah.sui:	pos: (-67000, 0, -9000)
./colorado.sui:	pos: (-43000, 0, -2000)
./oklahoma.sui:	pos: (-6000, 0, 17000)
./oregon.sui:	pos: (-97000, 0, -42000)
./washington.sui:	pos: (-93000, 0, -62500)
./kansas.sui:	pos: (-10000, 0, 3000)
./nevada.sui:	pos: (-90000, 0, -14000)
./idaho.sui:	pos: (-76000, 0, -41000)
./arizona.sui:	pos: (-72000, 0, 22000)
./new_mexico.sui:	pos: (-45000, 0, 24000)
./california.sui:	pos: (-100000, 0, 4000)
./texas.sui:	pos: (-15000, 0, 44000)
./wyoming.sui:	pos: (-47500, 0, -25000)
./arkansas.sui:	pos: (17000, 0, 20000)
./montana.sui:	pos: (-52000, 0, -48000)
./nebraska.sui:	pos: (-15000, 0, -15000)
```
* City position helps determine where a city's label in the world map, and also defines some bounding boxes that trigger a city's discovery. 
* Ferry (and FerryConnection) position is used to draw 2-D bezier curves on the map. The only useful elevation data for a ferry is probably in the FerryItem, which is associated with a physical thing on the map (note: there's a difference between Ferry and FerryItem... I could do a better job of making that distinction clear 🫠) .

I am doubtful z-values for these three things will be useful to ETS2LA (but I could be wrong!), because those three objects don't represent anything physical.

### Should you be using z-values from nodes, instead?

I am also doubtful of adding z-values to the BaseItem type (and I am now thinking the BaseItem doesn't even need x/y-values...)

The source of truth for a map item's position in the physical world (like a Model item, which is what's used to render building footprints in the map) is the node(s) associated with that map item. It's what's used during rendering/GeoJSON-creation. 

Taking a look at a random Model item and it's associated node:

image

The x/y values differ. The code that converts a Model item into a GeoJSON polygon uses the node's x/y as the source of truth:

https://github.com/truckermudgeon/maps/blob/613ddc6ddaf3fc23ff439ed75e1e947ed0386209/packages/clis/generator/geo-json.ts#L706-L710

Depending on how you're using the info, you may be better off using z-values in nodes, instead.

Tumppi066:
> ### How will z-values be used for Cities, Countries, and Ferries 🤔?

No clue tbh, I just added Z values to everything I could find.

If they don't make sense then they can just be removed, I don't think we'll do anything with the city and country positions.

And yup I do know the difference between the base ones and the items, I've also done the same xD

TLDR: We don't use them, they can be removed. I'm not at a PC that can run the parser this weekend so I'll update the PR on Monday if you don't merge your branches before that 👍

truckermudgeon:
FYI:
* 899e818701c47323560b2e7819ad168fcb5a57af adds z-values to prefab nodes and navCurves
* cbdcf7339c3caee2d1de4b9f6cbf0852bb3330e3 adds shoulderSpace values

Also added RoadLook::name in 0dcb053cc195262198b24fb2d3675a4d361902a0.

Lastly:

> and I am now thinking the BaseItem doesn't even need x/y-values...

I did some checking and it turns out there's at least one bug caused by using Item x/y values instead of Node x/y values... viewpoint icons aren't being rendered correctly 😕.

| Game (Thompson Falls, MT) | `maps` ([Thompson Falls, MT](https://truckermudgeon.github.io/#12.88/47.63993/-115.2849)) |
|----|----|
| image | image |

In the `maps` screenshot: The viewpoint icon is rendered with item coords (incorrect); the blue pin is rendered with node coords (correct). Will raise a PR to fix that.

Tumppi066:
Your version now has everything mine did, so this PR is closed 👍

PR #18, opened 2024-10-06
parser: parse ats achievements data

truckermudgeon:
This PR adds `parser` support for ATS Achievements-related data, writing out 5 new JSON files:

* `achievements`: a JSON-ified, reduced version of `def/achievements.sii`.
  * Only includes data for Achievements that I consider mappable. The set of mappable Achievements may grow in the future (e.g., to include Cargo-based Achievements like "Cotton Bloom"
  * Only includes data for ATS Achievements. Support for ETS2 Achievements will be added in a separate PR.
* `cutscenes`: a list of CutsceneItems. Needed to support Achievements related to viewpoints, e.g., "The Director"
* `routes`: a list of Special Transport routes
* `trajectories`: a list of TrajectoryItems with the `offer_point` checkpoint, which denote the starting point of a Special Transport route
* `triggers`: a list of Trigger items, which, among other things, holds information about trigger-based Achievements like "Start Your Engine!"

### DIFF
--- a/packages/clis/generator/mapped-data.ts
+++ b/packages/clis/generator/mapped-data.ts
@@ -30,7 +30,13 @@ import process from 'process';
 import { logger } from './logger';
 import { readArrayFile } from './read-array-file';
 
-const MapDataKeys: Record<keyof MapData, void> = {
+const MapDataKeys: Record<
+  Exclude<
+    keyof MapData,
+    'trajectories' | 'triggers' | 'cutscenes' | 'achievements' | 'routes'
+  >,
+  void
+> = {
   cities: undefined,
   companies: undefined,
   companyDefs: undefined,
@@ -57,7 +63,13 @@ export type MappedData = Omit<
   {
     [K in keyof MapData]: Map<string, MapData[K][0]>;
   },
-  'pois' | 'elevation'
+  | 'pois'
+  | 'elevation'
+  | 'trajectories'
+  | 'triggers'
+  | 'cutscenes'
+  | 'achievements'
+  | 'routes'
 > & {
   pois: MapData['pois'];
   elevation: MapData['elevation'];
--- a/packages/clis/parser/game-files/map-files-parser.ts
+++ b/packages/clis/parser/game-files/map-files-parser.ts
@@ -359,11 +359,6 @@ function convertSiiToJson<T>(
   }
 
   const json = jsonConverter.convert(res.cst);
-  if (schema == null) {
-    console.log(JSON.stringify(json, null, 2));
-    throw new Error();
-  }
-
   const validate = ajv.compile(schema);
   if (validate(json)) {
     return json;
--- a/packages/clis/parser/game-files/map-files-parser.ts
+++ b/packages/clis/parser/game-files/map-files-parser.ts
@@ -9,6 +9,7 @@ import {
 } from '@truckermudgeon/map/constants';
 import { toMapPosition } from '@truckermudgeon/map/prefabs';
 import type {
+  Achievement,
   Building,
   City,
   CityArea,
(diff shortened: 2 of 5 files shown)


packages/clis/generator/mapped-data.ts
```diff
@@ -30,7 +30,13 @@ import process from 'process';
 import { logger } from './logger';
 import { readArrayFile } from './read-array-file';
 
-const MapDataKeys: Record<keyof MapData, void> = {
+const MapDataKeys: Record<
+  Exclude<
+    keyof MapData,
+    'trajectories' | 'triggers' | 'cutscenes' | 'achievements' | 'routes'
```
truckermudgeon, line 36: Temporary exclusion list, just so code still compiles. Will be removed in follow-up `generator` PR.

packages/clis/generator/mapped-data.ts
```diff
@@ -57,7 +63,13 @@ export type MappedData = Omit<
...
     [K in keyof MapData]: Map<string, MapData[K][0]>;
   },
-  'pois' | 'elevation'
+  | 'pois'
+  | 'elevation'
+  | 'trajectories'
+  | 'triggers'
+  | 'cutscenes'
+  | 'achievements'
+  | 'routes'
```
truckermudgeon, line 72: Ditto

packages/clis/parser/game-files/map-files-parser.ts
```diff
@@ -329,6 +359,11 @@ function convertSiiToJson<T>(
   }
 
   const json = jsonConverter.convert(res.cst);
+  if (schema == null) {
+    console.log(JSON.stringify(json, null, 2));
+    throw new Error();
+  }
+
```
truckermudgeon:
```suggestion
```

packages/clis/parser/game-files/map-files-parser.ts
```diff
@@ -674,6 +709,147 @@ function processRoadLookJson(obj: RoadLookSii): Map<string, RoadLook> {
   );
 }
 
+function processAchievementsJson(
```
truckermudgeon, line 707: Note to self: split up these huge files 😬

packages/clis/parser/game-files/sector-parser.ts
```diff
@@ -410,8 +411,6 @@ const SimpleItemStruct = {
     vegetationSpheres,
     quadInfo,
   },
-  // useful for routing?
-  // https://modding.scssoft.com/wiki/New_Editor_Features_info_-_old_%2B_1.47
```
truckermudgeon, line 414: I thought it was... until I learned "trajectory" is used for special transport deliveries 🫠

packages/libs/map/types.ts
```diff
@@ -241,16 +298,19 @@ export type CompanyItem = BaseItem &
...
+    actionStringParams: readonly string[];
   }>;
 
 export type Trigger = BaseItem &
   Readonly<{
     type: ItemType.Trigger;
     dlcGuard: number;
-    actionTokens: readonly string[];
+    // [action token, params] tuples
+    actions: [string, readonly string[]][];
```
truckermudgeon, line 313: I originally wrote this as `Map<string, string[]>`. But then `JSON.serialize`ing Triggers resulted in empty objects being serialized 😕 

I don't think we're at the point where we need special serialize/deserialize routines yet... but maybe in the future.

packages/libs/map/types.ts
```diff
@@ -445,6 +511,11 @@ export type ContourFeature = GeoJSON.Feature<
   { elevation: number }
 >;
 
+export type AchievementFeature = GeoJSON.Feature<
+  GeoJSON.Polygon | GeoJSON.Point,
+  { name: string; dlcGuard: number }
+>;
+
```
truckermudgeon: This should be done in the `generator` PR.

packages/libs/map/types.ts
```diff
@@ -455,6 +526,7 @@ export type AtsMapGeoJsonFeature =
   | PoiFeature
   | FootprintFeature
   | ContourFeature
+  | AchievementFeature
```
truckermudgeon: Ditto.

PR #30, opened 2024-12-18
Skip parsing files related to the Winterland event

nautofon:
I can’t test this on ETS2 because I don’t have that game.

I don’t expect this patch will break anything, but the regex may need additional name components to skip. Either way, you’ll probably want to test this on ETS2 before merging.

### DIFF
--- a/packages/clis/parser/game-files/def-parser.ts
+++ b/packages/clis/parser/game-files/def-parser.ts
@@ -96,6 +96,9 @@ export function parseDefFiles(entries: Entries, application: 'ats' | 'eut2') {
     if (!/^(city|country|company|ferry)\./.test(f) || !f.endsWith('.sii')) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


truckermudgeon:
Did some testing and this works on ETS2, too. Thanks for this!

As an alternative, I briefly considered loosening the schema constraints to allow for the Winterland edge cases... but came to the conclusion it's not worth it, because I still don't know where the road and prefab data for Winterland lives, and so the generated map files end up being devoid of Winterland data, anyway 😕

nautofon:
Thanks! ☺️ I’ve started hacking this a little bit, and it’s just a lot easier when the parser doesn’t croak to begin with.

By the way, I’m not exactly sure what data you’d need to add Winterland, but I think you’d find some of it by grepping for these paths:
```
map/event2023/
map/usa.external.sii
prefab/corners/xmas_
_xmas.pmd
road_template/xmas
```

PR #31, opened 2024-12-21
Parse mileage targets and write them to an intermediate JSON array file

nautofon:
Adding a way to read the contents of `mileage_targets.sii` and resolve the node UIDs referenced in it will make it easier for me to update the scenery towns for the next DLC releases. Beyond that, it *might* be a stepping stone towards partially automating the process of updating scenery towns.

This patch is more or less my first foray into TypeScript and Node.js. As such, a thorough review might be appropriate. I can tell you up front that there are several things I'm not too happy with:
* As before, I didn't test this change with ETS2.
* [`nodeUid` in `MileageTarget`](https://github.com/nautofon/maps-nodejs/blob/d59dfe86d54ce69b890b94837ce03d4bda8bb02b/packages/libs/map/types.ts#L111) is declared as `string`. I think `bigint` would make more sense, but couldn't get it working right.
* There are no automated tests to verify the intermediate JSON file is created correctly.

Please don't feel rushed with this or anything, I'm not in a hurry here. After all, it's Christmas. ☺️

### DIFF
--- a/packages/clis/parser/game-files/def-parser.ts
+++ b/packages/clis/parser/game-files/def-parser.ts
@@ -9,6 +9,7 @@ import type {
   Ferry,
   FerryConnection,
   LaneSpeedClass,
+  MileageTarget,
   ModelDescription,
   PrefabDescription,
   RoadLook,
@@ -30,6 +31,7 @@ import type {
   CountrySii,
   Ets2AchievementsSii,
   FerrySii,
+  MileageTargetsSii,
   ModelSii,
   PrefabSii,
   RoadLookSii,
@@ -46,6 +48,7 @@ import {
   Ets2AchievementsSiiSchema,
   FerryConnectionSiiSchema,
   FerrySiiSchema,
+  MileageTargetsSiiSchema,
   ModelSiiSchema,
   PrefabSiiSchema,
   RoadLookSiiSchema,
@@ -207,6 +210,15 @@ export function parseDefFiles(entries: Entries, application: 'ats' | 'eut2') {
   logger.info('parsed', models.size, 'building models');
   logger.info('parsed', vegetation.size, 'vegetation models');
 
+  const mileageTargets: Map<string, MileageTarget> = processMileageTargetJson(
+    convertSiiToJson(
+      'def/sign/mileage_targets.sii',
+      entries,
+      MileageTargetsSiiSchema,
+    ),
+  );
+  logger.info('parsed', mileageTargets.size, 'mileage targets');
+
   const defPhotoAlbum = Preconditions.checkExists(
     entries.directories.get('def/photo_album'),
   );
@@ -273,6 +285,7 @@ export function parseDefFiles(entries: Entries, application: 'ats' | 'eut2') {
     roadLooks,
     models,
     vegetation,
+    mileageTargets,
     viewpoints,
   };
 }
@@ -908,4 +921,40 @@ function processRouteJson(obj: RouteSii): Map<string, Route> {
     routes.set(routeKey, route);
   }
   return routes;
+}
+
+function processMileageTargetJson(
+  obj: MileageTargetsSii,
+): Map<string, MileageTarget> {
+  const mileageTargets = new Map<string, MileageTarget>();
+  for (const [key, rawTarget] of Object.entries(obj.mileageTarget)) {
+    const token = assertExists(key.split('.')[1]);
+    let target: MileageTarget = {
+      token: token,
+      editorName: rawTarget.editorName,
+      defaultName: rawTarget.defaultName,
+      nameVariants: Array.isArray(rawTarget.names) ? rawTarget.names : [],
+      distanceOffset: rawTarget.distanceOffset,
+    };
+    // Some mileage targets are specified with a position, which we can use
+    // directly. For other mileage targets, a node uid is given, which we'll
+    // have to try to resolve later using data from map sector files.
+    if (rawTarget.position.every(v => v != null)) {
+      // SCS coordinates: easting, up, southing
+      const [x, , y] = rawTarget.position.map(v => Math.round(v * 100) / 100);
+      target = { ...target, x, y };
+    } else if (rawTarget.nodeUid != null) {
+      target = { ...target, nodeUid: rawTarget.nodeUid };
+    } else {
+      // A total lack of position information is rare, but it can happen for
+      // test data or unreleased DLC. Either way, nothing we can do about it.
+      logger.debug('skipping mileage target (no position, nor uid)', token);
+      continue;
+    }
+    if (rawTarget.searchRadius >= 0) {
+      target = { ...target, searchRadius: rawTarget.searchRadius };
+    }
+    mileageTargets.set(token, target);
+  }
+  return mileageTargets;
 }
--- a/packages/clis/parser/game-files/map-files-parser.ts
+++ b/packages/clis/parser/game-files/map-files-parser.ts
@@ -925,6 +925,25 @@ function postProcess(
     'roads possibly split by terrains, buildings, or curves',
   );
 
+  // Augment mileage targets from defs with position info from sectors.
+  for (const [token, target] of defData.mileageTargets) {
+    if ((target.x && target.y) || !target.nodeUid) {
+      continue;
+    }
+    const { nodeUid, ...targetWithoutNodeUid } = target;
+    const node = nodesByUid.get(nodeUid);
+    if (node) {
+      defData.mileageTargets.set(token, {
+        ...targetWithoutNodeUid,
+        x: Math.round(node.x * 100) / 100, // easting
+        y: Math.round(node.y * 100) / 100, // southing
+      });
+      logger.trace('node', nodeUid, 'found for mileage target', token);
+    } else {
+      logger.debug('node', nodeUid, 'not found for mileage target', token);
+    }
+  }
+
   logger.info(elevationNodeUids.size, 'elevation nodes');
   const referencedNodes: Node[] = [];
   for (const uid of referencedNodeUids) {
@@ -962,6 +981,7 @@ function postProcess(
       modelDescriptions: valuesWithTokens(defData.models),
       achievements: valuesWithTokens(defData.achievements),
       routes: valuesWithTokens(defData.routes),
+      mileageTargets: valuesWithTokens(defData.mileageTargets),
     },
     icons,
   };
@@ -980,6 +1000,7 @@ function toDefData(
     modelDescriptions: valuesWithTokens(defData.models),
     achievements: valuesWithTokens(defData.achievements),
     routes: valuesWithTokens(defData.routes),
+    mileageTargets: valuesWithTokens(defData.mileageTargets),
   };
 }
 
--- a/packages/clis/parser/game-files/sii-parser.ts
+++ b/packages/clis/parser/game-files/sii-parser.ts
@@ -39,6 +39,11 @@ const NumberLiteral = createToken({
   longer_alt: [HexLiteral, Property],
   pattern: /[-+]?(0|[1-9]\d*)(\.\d+)?([eE][+-]?\d+)?/,
 });
+const BinaryFloat = createToken({
+  name: 'BinaryFloat',
+  // Hexadecimal representation of IEEE 754 binary32 floats, big-endian.
+  pattern: /&[0-9a-fA-F]{8}/,
+});
 const WhiteSpace = createToken({
   name: 'WhiteSpace',
   pattern: /\s+/,
@@ -63,6 +68,7 @@ const allTokens = [
   AtInclude,
   StringLiteral,
   NumberLiteral,
+  BinaryFloat,
   HexLiteral,
   Property,
   WhiteSpace,
@@ -86,7 +92,11 @@ class SiiParser extends CstParser {
     this.CONSUME(LParen);
     this.AT_LEAST_ONE_SEP({
       SEP: Comma,
-      DEF: () => this.CONSUME(NumberLiteral),
+      DEF: () =>
+        this.OR([
+          { ALT: () => this.CONSUME(NumberLiteral) },
+          { ALT: () => this.CONSUME(BinaryFloat) },
+        ]),
     });
     this.CONSUME(RParen);
   });
@@ -102,6 +112,7 @@ class SiiParser extends CstParser {
     this.OR([
       { ALT: () => this.CONSUME(StringLiteral) },
       { ALT: () => this.CONSUME(NumberLiteral) },
+      { ALT: () => this.CONSUME(BinaryFloat) },
       { ALT: () => this.CONSUME(HexLiteral) },
       { ALT: () => this.CONSUME(Property) },
       { ALT: () => this.SUBRULE(this.numberTuple) },
--- a/packages/clis/parser/game-files/sii-parser.ts
+++ b/packages/clis/parser/game-files/sii-parser.ts
@@ -24,6 +24,10 @@ const Property = createToken({
   // The negative lookahead for `x` is to ensure HexLiterals can be parsed.
   pattern: /([a-zA-Z0-9_.]+)|((0(?!x))?[1-9]+[a-z_.]+[a-z0-9_.]+)/,
 });
+const NilLiteral = createToken({
+  name: 'Nil',
+  pattern: /nil/,
+});
 const StringLiteral = createToken({
   name: 'String',
   // Can't simply use /"[^"]*"/, because we might have string literals with
@@ -66,6 +70,7 @@ const allTokens = [
   Comma,
   Colon,
   AtInclude,
+  NilLiteral,
   StringLiteral,
   NumberLiteral,
   BinaryFloat,
@@ -110,6 +115,7 @@ class SiiParser extends CstParser {
   });
   readonly objectPropertyValue = this.RULE('objectPropertyValue', () => {
     this.OR([
+      { ALT: () => this.CONSUME(NilLiteral) },
       { ALT: () => this.CONSUME(StringLiteral) },
       { ALT: () => this.CONSUME(NumberLiteral) },
       { ALT: () => this.CONSUME(BinaryFloat) },
--- a/packages/clis/parser/game-files/sii-schemas.ts
+++ b/packages/clis/parser/game-files/sii-schemas.ts
@@ -30,6 +30,7 @@ ajv.addKeyword({
 const integer: JSONSchemaType<number> = { type: 'integer' } as const;
 const number: JSONSchemaType<number> = { type: 'number' } as const;
 const string: JSONSchemaType<string> = { type: 'string' } as const;
+const null_: JSONSchemaType<null> = { type: 'null', nullable: true } as const;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
 const bigint: JSONSchemaType<bigint> = { dataType: 'bigint' } as any;
 
@@ -37,6 +38,12 @@ type Tuple<T, N extends number, R extends T[] = []> = R['length'] extends N
   ? R
   : Tuple<T, N, [T, ...R]>;
 
+const constInteger = <T extends number>(n: T) =>
+  ({
+    type: 'integer',
+    minimum: n,
+    maximum: n,
+  }) as JSONSchemaType<T>;
 const arrayOf = <T>(
   items: T,
   size: { minItems?: number; maxItems?: number } = { minItems: 1 },
@@ -540,6 +547,53 @@ export const CitySiiSchema: JSONSchemaType<CitySii> = object(
   },
   [],
 );
+
+export interface MileageTargetsSii {
+  mileageTarget: Record<
+    string,
+    {
+      editorName: string;
+      defaultName: string;
+      // The number zero as `names` signifies an empty array of strings.
+      names: 0 | string[];
+      distanceOffset: number;
+      nodeUid?: bigint;
+      position: [number, number, number] | [null, null, null];
+      searchRadius: number;
+    }
+  >;
+}
+export const MileageTargetsSiiSchema: JSONSchemaType<MileageTargetsSii> =
+  object(
+    {
+      mileageTarget: patternRecord(
+        /^mileage\.[0-9a-z_]{1,12}$/,
+        {
+          editorName: string,
+          defaultName: string,
+          names: { anyOf: [constInteger(0), stringArray] },
+          distanceOffset: number,
+          // `nodeUid` is optional, but the `nullable` combinator can't be used
+          // here because it's not compatible with the custom `bigint`
+          // descriptor. instead, cast it to the type that `nullable` produces.
+          nodeUid: bigint as typeof bigint & { nullable: true },
+          position: {
+            anyOf: [fixedLengthArray(number, 3), fixedLengthArray(null_, 3)],
+          },
+          searchRadius: number,
+        },
+        [
+          'editorName',
+          'defaultName',
+          'names',
+          'distanceOffset',
+          'position',
+          'searchRadius',
+        ],
+      ),
+    },
+    [],
+  );
 
 export interface ViewpointsSii {
   photoAlbumItem: Record<
--- a/packages/clis/parser/game-files/sii-visitors.ts
+++ b/packages/clis/parser/game-files/sii-visitors.ts
@@ -76,6 +76,12 @@ class JsonConverterVisitor extends getSiiVisitorClass<
       json.value = quotedStringToString(children.String[0].image);
     } else if (children.NumberLiteral) {
       json.value = stringToNumber(children.NumberLiteral[0].image);
+      if (
+        (json.value as number) > Number.MAX_SAFE_INTEGER &&
+        /^\d+$/.test(children.NumberLiteral[0].image)
+      ) {
+        json.value = BigInt(children.NumberLiteral[0].image);
+      }
     } else if (children.BinaryFloat) {
       json.value = stringToFloat(children.BinaryFloat[0].image);
     } else if (children.HexLiteral) {
--- a/packages/clis/parser/game-files/sii-visitors.ts
+++ b/packages/clis/parser/game-files/sii-visitors.ts
@@ -76,15 +76,24 @@ class JsonConverterVisitor extends getSiiVisitorClass<
       json.value = quotedStringToString(children.String[0].image);
     } else if (children.NumberLiteral) {
       json.value = stringToNumber(children.NumberLiteral[0].image);
+    } else if (children.BinaryFloat) {
+      json.value = stringToFloat(children.BinaryFloat[0].image);
     } else if (children.HexLiteral) {
       const bigInt = BigInt(children.HexLiteral[0].image);
       json.value = bigInt <= Number.MAX_SAFE_INTEGER ? Number(bigInt) : bigInt;
     } else if (children.Property) {
       json.value = children.Property[0].image;
     } else if (children.numberTuple) {
-      json.value = children.numberTuple[0].children.NumberLiteral.map(l =>
-        stringToNumber(l.image),
+      const arr: NumberTupleChild[] = [];
+      children.numberTuple[0].children.NumberLiteral?.forEach(l =>
+        arr.push({ value: stringToNumber(l.image), offset: l.startOffset }),
       );
+      children.numberTuple[0].children.BinaryFloat?.forEach(l =>
+        arr.push({ value: stringToFloat(l.image), offset: l.startOffset }),
+      );
+      // Regular numbers and binary floats may be mixed in any order.
+      // Sort by the offset to get the original order from the SII file.
+      json.value = arr.sort((a, b) => a.offset - b.offset).map(x => x.value);
     } else if (children.numberAuxTuple) {
       json.value = children.numberAuxTuple[0].children.NumberLiteral.map(l =>
         stringToNumber(l.image),
@@ -99,6 +108,11 @@ class JsonConverterVisitor extends getSiiVisitorClass<
   override includeDirective(children: IncludeDirectiveCstChildren) {
     logger.warn('ignoring @include directive', children.String[0].image);
   }
+}
+
+interface NumberTupleChild {
+  value: number | null;
+  offset: number;
 }
 
 class IncludeDirectiveCollector extends getSiiVisitorClass<string[]>() {
@@ -143,4 +157,18 @@ function stringToNumber(str: string) {
     throw new Error('could not parse number: ' + str);
   }
   return num;
+}
+
+function stringToFloat(str: string): number | null {
+  const binaryInt = parseInt(str.substring(1), 16);
+  if (isNaN(binaryInt)) {
+    throw new Error('could not parse binary float: ' + str);
+  }
+  if (binaryInt == 0x7f7fffff) {
+    // binary32 float max finite value, used as a "no data" marker
+    return null;
+  }
+  const data = new DataView(new ArrayBuffer(4));
+  data.setUint32(0, binaryInt);
+  return data.getFloat32(0);
 }
--- a/packages/clis/parser/game-files/sii-visitors.ts
+++ b/packages/clis/parser/game-files/sii-visitors.ts
@@ -62,7 +62,7 @@ class JsonConverterVisitor extends getSiiVisitorClass<
         } else {
           const tempWrapper = { value: undefined };
           this.visit(p.objectPropertyValue, tempWrapper);
-          obj[propKey] = assertExists(tempWrapper.value);
+          obj[propKey] = tempWrapper.value;
         }
       }
     }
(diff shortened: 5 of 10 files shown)


packages/clis/parser/types/sii-visitor.d.ts
```diff
@@ -28,7 +28,8 @@ export interface NumberTupleCstNode extends CstNode {
 
 export type NumberTupleCstChildren = {
   LParen: IToken[];
-  NumberLiteral: IToken[];
+  NumberLiteral?: IToken[];
```
truckermudgeon, line 35: It's horribly undocumented, but this file is actually generated using a `chevrotain` helper function... and the way I generate it is by [running a unit test](https://github.com/truckermudgeon/maps/blob/d59dfe86d54ce69b890b94837ce03d4bda8bb02b/packages/clis/parser/game-files/tests/sii-parser.test.ts#L254-L258) and copy-pasting the console output into this file, because I was too lazy to do it in a more-proper way 🙃.

packages/libs/map/types.ts
```diff
@@ -102,6 +102,18 @@ export type Ferry = Readonly<{
   connections: FerryConnection[];
 }>;
 
+export type MileageTarget = Readonly<{
+  token: string;
+  editorName: string;
+  defaultName: string;
+  nameVariants: string[];
+  distanceOffset: number;
+  nodeUid?: string;
```
truckermudgeon: > [nodeUid in MileageTarget](https://github.com/nautofon/maps-nodejs/blob/d59dfe86d54ce69b890b94837ce03d4bda8bb02b/packages/libs/map/types.ts#L111) is declared as `string`. I think bigint would make more sense, but couldn't get it working right.

I agree `bigint` would make more sense, too.

I went on a bit of a typing journey, and think I found a way... it involved:
* parsing `nil` values in the `.sii` files as a first-class Nil token (instead of as a generic string-y property/token value)
* resorting to a small cast in `MileageTargetsSiiSchema`, because of some weirdness with nullable JSON Schema values.

See `.diff` file in review comment 🙂

packages/libs/map/types.ts
```diff
@@ -102,6 +102,18 @@ export type Ferry = Readonly<{
...
+  easting?: number;
+  southing?: number;
```
truckermudgeon: If you're not opposed, could you rename these to `x` and `y`, for consistency with the other types containing position information?

packages/clis/parser/game-files/sii-visitors.ts
```diff
@@ -101,6 +116,11 @@ class JsonConverterVisitor extends getSiiVisitorClass<
   }
 }
 
+interface NumberTupleChild {
+  value: number | null;
```
truckermudgeon, line 122: Nit: for type declarations, could you use `undefined` instead of `null`?

The convention that I've been following (which I know is a bit weird, but it's what I grew up with 🙃) is to prefer `undefined` in type and value declarations, but to prefer `null` and non-strict equality checks, e.g.:

```ts
type Foo = { type: 'foo', value: number | undefined };
const aFoo: Foo = { type: 'foo', value: 5 };
const anotherFoo: Foo = { type: 'foo', value: undefined };

function maybeReturnFoo(f: Foo): Foo | undefined {
  if (f.value == null) {
    return undefined; // explicitly return `undefined`
    // it'd also be ok to do:
    //   return;
    // because the return value is implicitly `undefined`, thanks to signature.
    // but i'd still prefer the first, explicit way.
  }
  assert(f.value != null);
  return f;
}
```

packages/clis/parser/game-files/sii-visitors.ts
```diff
@@ -74,17 +74,32 @@ class JsonConverterVisitor extends getSiiVisitorClass<
   ): void {
     if (children.String) {
       json.value = quotedStringToString(children.String[0].image);
+    } else if (children.BinaryFloat) {
+      json.value = stringToFloat(children.BinaryFloat[0].image);
```
truckermudgeon: Nit: could you re-arrange to follow the order of alternates in `sii-parser.ts`' `SiiParser::objectPropertyValue`?

https://github.com/truckermudgeon/maps/blob/d59dfe86d54ce69b890b94837ce03d4bda8bb02b/packages/clis/parser/game-files/sii-parser.ts#L111-L121

The BinaryFloat branch should come after the NumberLiteral branch.

packages/clis/parser/game-files/sii-parser.ts
```diff
@@ -30,6 +30,11 @@ const StringLiteral = createToken({
   // escaped double-quotes, e.g.: "this is \"fine\"".
   pattern: /"(?:[^"\\]|\\.)*"/,
 });
+const BinaryFloat = createToken({
+  name: 'BinaryFloat',
+  pattern: /&[0-9a-fA-F]{8}/,
+  // Hexadecimal representation of IEEE 754 binary32 floats, big-endian.
```
truckermudgeon: Nit: 
```suggestion
  // Hexadecimal representation of IEEE 754 binary32 floats, big-endian.
  pattern: /&[0-9a-fA-F]{8}/,
```

packages/clis/parser/game-files/def-parser.ts
```diff
@@ -909,3 +922,39 @@ function processRouteJson(obj: RouteSii): Map<string, Route> {
...
+    if (rawTarget.position[0] != null) {
+      assert(rawTarget.position[1] != null && rawTarget.position[2] != null);
+      const [e, , s] = rawTarget.position.map(c => Math.round(c * 100) / 100);
+      target = { ...target, easting: e, southing: s };
+    } else if (rawTarget.nodeUid != 'nil') {
+      target = { ...target, nodeUid: rawTarget.nodeUid };
+    } else {
+      // A total lack of position information is rare, but it can happen for
+      // test data or unreleased DLC. Either way, nothing we can do about it.
+      logger.debug('skipping mileage target (no position, nor uid)', token);
```
truckermudgeon, line 951: Optional: WDYT of making this a `warn`? ATS produces two warnings:

```
⚠ skipping mileage target (no position, nor uid) nv_sparks                                                                                                                                       
⚠ skipping mileage target (no position, nor uid) aaa_test  
```

and ETS2 produces a few more:

```
⚠ skipping mileage target (no position, nor uid) ru_sosnovy                                                                                                                                      
⚠ skipping mileage target (no position, nor uid) ro_orsova                                                                                                                                       
⚠ skipping mileage target (no position, nor uid) ro_buzau                                                                                                                                        
⚠ skipping mileage target (no position, nor uid) bg_montana                                                                                                                                      
⚠ skipping mileage target (no position, nor uid) tr_kesan                                                                                                                                        
⚠ skipping mileage target (no position, nor uid) it_grosseto                                                                                                                                     
⚠ skipping mileage target (no position, nor uid) fr_stquentin                                                                                                                                    
⚠ skipping mileage target (no position, nor uid) bg_vidin                                                                                                                                        
⚠ skipping mileage target (no position, nor uid) ru_mamonovo                                                                                                                                     
⚠ skipping mileage target (no position, nor uid) lt_kybartai                                                                                                                                     
⚠ skipping mileage target (no position, nor uid) ru_torfyanov                                                                                                                                    
⚠ skipping mileage target (no position, nor uid) ro_calafat                                                                                                                                      
⚠ skipping mileage target (no position, nor uid) ru_moshok                                                                                                                                       
⚠ skipping mileage target (no position, nor uid) pl_korczowa    
```

packages/clis/parser/game-files/map-files-parser.ts
```diff
@@ -925,6 +925,25 @@ function postProcess(
...
+    const node = nodesByUid.get(nodeUid);
+    if (node) {
+      defData.mileageTargets.set(token, {
+        ...target,
+        easting: Math.round(node.x * 100) / 100,
+        southing: Math.round(node.y * 100) / 100,
+      });
+      logger.trace('node', nodeUid, 'found for mileage target', token);
+    } else {
+      logger.debug('node', nodeUid, 'not found for mileage target', token);
```
truckermudgeon, line 943: Similar optional question as above: WDYT of making this a `warn`?

(🤔 maybe the real questions are: what's the difference between a warning-level and a debug-level message, and should the `parser` command allow users to pass in a desired log level)

packages/clis/parser/game-files/map-files-parser.ts
```diff
@@ -925,6 +925,25 @@ function postProcess(
...
+  // Augment mileage targets from defs with position info from sectors.
+  for (const [token, target] of defData.mileageTargets) {
+    if ((target.easting && target.southing) || !target.nodeUid) {
+      continue;
+    }
+    const nodeUid = BigInt(target.nodeUid);
+    const node = nodesByUid.get(nodeUid);
+    if (node) {
+      defData.mileageTargets.set(token, {
+        ...target,
```
truckermudgeon: It looks like we're serializing both easting/southing _and_ `nodeUid` here.

If `nodeUid` might be used for future analysis, then it should probably be added to the `referencedNodeUids` collection, because that collection is what determines which Nodes are serialized in `{usa,europe}-nodes.json`.

If it's not useful / it's just redundant info, then maybe the `MileageTarget` type shouldn't have a `nodeUid` property? It can be excluded from serialization by doing something like:

```ts
const { nodeUid, ...targetWithoutNodeUid } = target;
defData.mileageTargets.set(token, {
  ...targetWithoutNodeUid,
  // ...etc

File: packages/clis/parser/game-files/sii-visitors.ts [hunk 10]
```diff
@@ -74,17 +74,32 @@ class JsonConverterVisitor extends getSiiVisitorClass<
...
-        stringToNumber(l.image),
+      const arr: NumberTupleChild[] = [];
+      children.numberTuple[0].children.NumberLiteral?.forEach(l =>
+        arr.push({ value: stringToNumber(l.image), offset: l.startOffset }),
+      );
+      children.numberTuple[0].children.BinaryFloat?.forEach(l =>
+        arr.push({ value: stringToFloat(l.image), offset: l.startOffset }),
       );
+      // Regular numbers and binary floats may be mixed in any order.
+      // Sort by the offset to get the original order from the SII file.
```
truckermudgeon, line 103: 👍

packages/clis/parser/game-files/sii-visitors.ts
```diff
@@ -144,3 +164,17 @@ function stringToNumber(str: string) {
...
 }
+
+function stringToFloat(str: string) {
+  const binaryInt = parseInt(str.substring(1), 16);
+  if (isNaN(binaryInt)) {
+    throw new Error('could not parse binary float: ' + str);
+  }
+  if (binaryInt == 0x7f7fffff) {
+    // binary32 float max finite value, used as a "no data" marker
+    return null;
```
truckermudgeon:
```suggestion
function stringToFloat(str: string): number | undefined {
  const binaryInt = parseInt(str.substring(1), 16);
  if (isNaN(binaryInt)) {
    throw new Error('could not parse binary float: ' + str);
  }
  if (binaryInt == 0x7f7fffff) {
    // binary32 float max finite value, used as a "no data" marker
    return undefined;
```

packages/clis/parser/game-files/def-parser.ts
```diff
@@ -909,3 +922,39 @@ function processRouteJson(obj: RouteSii): Map<string, Route> {
...
+      editorName: rawTarget.editorName,
+      defaultName: rawTarget.defaultName,
+      nameVariants: Array.isArray(rawTarget.names) ? rawTarget.names : [],
+      distanceOffset: rawTarget.distanceOffset,
+    };
+    // Some mileage targets are specified with a position, which we can use
+    // directly. For other mileage targets, a node uid is given, which we'll
+    // have to try to resolve later using data from map sector files.
+    if (rawTarget.position[0] != null) {
+      assert(rawTarget.position[1] != null && rawTarget.position[2] != null);
```
truckermudgeon: If the schema guarantees that this array is either all numbers or all undefineds (see diff file), then this can look like:

```suggestion
    if (rawTarget.position.every(v => v != null)) {
```

packages/clis/parser/game-files/sii-schemas.ts
```diff
@@ -541,6 +541,36 @@ export const CitySiiSchema: JSONSchemaType<CitySii> = object(
...
+export interface MileageTargetsSii {
+  mileageTarget: Record<
+    string,
+    {
+      editorName: string;
+      defaultName: string;
+      names: number | string[];
+      distanceOffset: number;
+      nodeUid: string;
+      position: [number, number, number];
```
truckermudgeon: Because of how the schema is declared as:

```ts
        position: fixedLengthArray(nullable(number), 3),
```

this should be typed as `[number | undefined, number | undefined, number | undefined]` to match.

But I think what we want is:

```
[number, number, number] | [undefined, undefined, undefined]
```

because AFAICT, `position` is only ever fully populated with valid data, or fully populated with the "no data" marker.

See `.diff` for changes related to this.

truckermudgeon (changes requested):
What a cool PR to see!

> This patch is more or less my first foray into TypeScript and Node.js.

👏 well done, especially for your first time! And _especially_ because there's a lot of idiosyncratic code and niche libraries to deal with (namely `chevrotain` for parsing, and `ajv` for JSON schema declaration and validation). Thank you for persevering 🙂 

My comments are largely nits. The only non-trivial comments I had are around the types  and schema validation of mileage target `position` and `nodeUid`. I had a go massaging both of them, and came up with:

<details>
<summary>Click to see <tt>.diff</tt></summary>

```diff
diff --git a/packages/clis/parser/game-files/def-parser.ts b/packages/clis/parser/game-files/def-parser.ts
index be88cbc..a795b9b 100644
--- a/packages/clis/parser/game-files/def-parser.ts
+++ b/packages/clis/parser/game-files/def-parser.ts
@@ -939,11 +939,10 @@ function processMileageTargetJson(
     // Some mileage targets are specified with a position, which we can use
     // directly. For other mileage targets, a node uid is given, which we'll
     // have to try to resolve later using data from map sector files.
-    if (rawTarget.position[0] != null) {
-      assert(rawTarget.position[1] != null && rawTarget.position[2] != null);
+    if (rawTarget.position.every(v => v != null)) {
       const [e, , s] = rawTarget.position.map(c => Math.round(c * 100) / 100);
       target = { ...target, easting: e, southing: s };
-    } else if (rawTarget.nodeUid != 'nil') {
+    } else if (rawTarget.nodeUid != null) {
       target = { ...target, nodeUid: rawTarget.nodeUid };
     } else {
       // A total lack of position information is rare, but it can happen for
diff --git a/packages/clis/parser/game-files/sii-parser.ts b/packages/clis/parser/game-files/sii-parser.ts
index 2230846..41af3c8 100644
--- a/packages/clis/parser/game-files/sii-parser.ts
+++ b/packages/clis/parser/game-files/sii-parser.ts
@@ -24,6 +24,10 @@ const Property = createToken({
   // The negative lookahead for `x` is to ensure HexLiterals can be parsed.
   pattern: /([a-zA-Z0-9_.]+)|((0(?!x))?[1-9]+[a-z_.]+[a-z0-9_.]+)/,
 });
+const NilLiteral = createToken({
+  name: 'Nil',
+  pattern: /nil/,
+});
 const StringLiteral = createToken({
   name: 'String',
   // Can't simply use /"[^"]*"/, because we might have string literals with
@@ -66,6 +70,7 @@ const allTokens = [
   Comma,
   Colon,
   AtInclude,
+  NilLiteral,
   StringLiteral,
   NumberLiteral,
   BinaryFloat,
@@ -110,6 +115,7 @@ class SiiParser extends CstParser {
   });
   readonly objectPropertyValue = this.RULE('objectPropertyValue', () => {
     this.OR([
+      { ALT: () => this.CONSUME(NilLiteral) },
       { ALT: () => this.CONSUME(StringLiteral) },
       { ALT: () => this.CONSUME(NumberLiteral) },
       { ALT: () => this.CONSUME(BinaryFloat) },
diff --git a/packages/clis/parser/game-files/sii-schemas.ts b/packages/clis/parser/game-files/sii-schemas.ts
index 0090bbc..a648db7 100644
--- a/packages/clis/parser/game-files/sii-schemas.ts
+++ b/packages/clis/parser/game-files/sii-schemas.ts
@@ -30,6 +30,7 @@ ajv.addKeyword({
 const integer: JSONSchemaType<number> = { type: 'integer' } as const;
 const number: JSONSchemaType<number> = { type: 'number' } as const;
 const string: JSONSchemaType<string> = { type: 'string' } as const;
+const null_: JSONSchemaType<null> = { type: 'null', nullable: true } as const;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
 const bigint: JSONSchemaType<bigint> = { dataType: 'bigint' } as any;
 
@@ -37,6 +38,12 @@ type Tuple<T, N extends number, R extends T[] = []> = R['length'] extends N
   ? R
   : Tuple<T, N, [T, ...R]>;
 
+const constInteger = <T extends number>(n: T) =>
+  ({
+    type: 'integer',
+    minimum: n,
+    maximum: n,
+  }) as JSONSchemaType<T>;
 const arrayOf = <T>(
   items: T,
   size: { minItems?: number; maxItems?: number } = { minItems: 1 },
@@ -547,10 +554,12 @@ export interface MileageTargetsSii {
     {
       editorName: string;
       defaultName: string;
-      names: number | string[];
+      // `names` is described as a number | string[] because the property is
+      // either 0 (describing an empty array), or a non-empty array of strings.
+      names: 0 | string[];
       distanceOffset: number;
-      nodeUid: string;
-      position: [number, number, number];
+      nodeUid?: bigint;
+      position: [number, number, number] | [undefined, undefined, undefined];
       searchRadius: number;
     }
   >;
@@ -558,15 +567,31 @@ export interface MileageTargetsSii {
 export const MileageTargetsSiiSchema: JSONSchemaType<MileageTargetsSii> =
   object(
     {
-      mileageTarget: patternRecord(/^mileage\.[0-9a-z_]{1,12}$/, {
-        editorName: string,
-        defaultName: string,
-        names: { anyOf: [number, stringArray] },
-        distanceOffset: number,
-        nodeUid: { anyOf: [bigint, string] },
-        position: fixedLengthArray(nullable(number), 3),
-        searchRadius: number,
-      }),
+      mileageTarget: patternRecord(
+        /^mileage\.[0-9a-z_]{1,12}$/,
+        {
+          editorName: string,
+          defaultName: string,
+          names: { anyOf: [constInteger(0), stringArray] },
+          distanceOffset: number,
+          // `nodeUid` is optional, but the `nullable` combinator can't be used
+          // here because it's not compatible with the custom `bigint`
+          // descriptor. instead, cast it to the type that `nullable` produces.
+          nodeUid: bigint as typeof bigint & { nullable: true },
+          position: {
+            anyOf: [fixedLengthArray(number, 3), fixedLengthArray(null_, 3)],
+          },
+          searchRadius: number,
+        },
+        [
+          'editorName',
+          'defaultName',
+          'names',
+          'distanceOffset',
+          'position',
+          'searchRadius',
+        ],
+      ),
     },
     [],
   );
diff --git a/packages/clis/parser/game-files/sii-visitors.ts b/packages/clis/parser/game-files/sii-visitors.ts
index c2e1d42..7449247 100644
--- a/packages/clis/parser/game-files/sii-visitors.ts
+++ b/packages/clis/parser/game-files/sii-visitors.ts
@@ -62,7 +62,7 @@ class JsonConverterVisitor extends getSiiVisitorClass<
         } else {
           const tempWrapper = { value: undefined };
           this.visit(p.objectPropertyValue, tempWrapper);
-          obj[propKey] = assertExists(tempWrapper.value);
+          obj[propKey] = tempWrapper.value;
         }
       }
     }
@@ -72,10 +72,10 @@ class JsonConverterVisitor extends getSiiVisitorClass<
     children: ObjectPropertyValueCstChildren,
     json: { value: unknown },
   ): void {
-    if (children.String) {
+    if (children.Nil) {
+      json.value = undefined;
+    } else if (children.String) {
       json.value = quotedStringToString(children.String[0].image);
-    } else if (children.BinaryFloat) {
-      json.value = stringToFloat(children.BinaryFloat[0].image);
     } else if (children.NumberLiteral) {
       json.value = stringToNumber(children.NumberLiteral[0].image);
       if (
@@ -84,6 +84,8 @@ class JsonConverterVisitor extends getSiiVisitorClass<
       ) {
         json.value = BigInt(children.NumberLiteral[0].image);
       }
+    } else if (children.BinaryFloat) {
+      json.value = stringToFloat(children.BinaryFloat[0].image);
     } else if (children.HexLiteral) {
       const bigInt = BigInt(children.HexLiteral[0].image);
       json.value = bigInt <= Number.MAX_SAFE_INTEGER ? Number(bigInt) : bigInt;
diff --git a/packages/clis/parser/game-files/tests/sii-visitors.test.ts b/packages/clis/parser/game-files/tests/sii-visitors.test.ts
index 8aba715..c8444db 100644
--- a/packages/clis/parser/game-files/tests/sii-visitors.test.ts
+++ b/packages/clis/parser/game-files/tests/sii-visitors.test.ts
@@ -72,7 +72,7 @@ mileage_target : mileage.ok_oklacity {
           imageAtlasIndices: 0,
           imageAtlasPaths: 0,
           names: ['Okla. City'],
-          nodeUid: 'nil',
+          nodeUid: undefined,
           position: [-7257.50341796875, 21.56877326965332, 19497.71875],
           searchRadius: 50,
           variants: ['okla_city'],
diff --git a/packages/clis/parser/types/sii-visitor.d.ts b/packages/clis/parser/types/sii-visitor.d.ts
index 9616be6..1f3b776 100644
--- a/packages/clis/parser/types/sii-visitor.d.ts
+++ b/packages/clis/parser/types/sii-visitor.d.ts
@@ -52,10 +52,11 @@ export interface ObjectPropertyValueCstNode extends CstNode {
 }
 
 export type ObjectPropertyValueCstChildren = {
+  Nil?: IToken[];
   String?: IToken[];
   NumberLiteral?: IToken[];
-  HexLiteral?: IToken[];
   BinaryFloat?: IToken[];
+  HexLiteral?: IToken[];
   Property?: IToken[];
   numberTuple?: NumberTupleCstNode[];
   numberAuxTuple?: NumberAuxTupleCstNode[];
diff --git a/packages/libs/map/types.ts b/packages/libs/map/types.ts
index ff92cce..3115472 100644
--- a/packages/libs/map/types.ts
+++ b/packages/libs/map/types.ts
@@ -108,7 +108,7 @@ export type MileageTarget = Readonly<{
   defaultName: string;
   nameVariants: string[];
   distanceOffset: number;
-  nodeUid?: string;
+  nodeUid?: bigint;
   easting?: number;
   southing?: number;
   searchRadius?: number;
```
</details>

Let me know what you think, and if you have any questions.

> Please don't feel rushed with this or anything, I'm not in a hurry here. After all, it's Christmas. ☺️

🎅 I'll consider this as a Christmas treat, haha. I was more than happy to review!

packages/clis/parser/types/sii-visitor.d.ts
truckermudgeon: This file can now be generated with an npm script (see 492b2fe). From the repo root:

```sh
npm run gen:types -w packages/clis/parser  
```

nautofon: Great!

How would you feel about additionally having a comment in the generated file? Something to let hapless noobs like myself know what's going on there. For example:

```patch
diff --git a/packages/clis/parser/game-files/sii-parser.ts b/packages/clis/parser/game-files/sii-parser.ts
index 0e450c7..357f33f 100644
--- a/packages/clis/parser/game-files/sii-parser.ts
+++ b/packages/clis/parser/game-files/sii-parser.ts
@@ -206 +206,7 @@ export function generateVisitorDts(): string {
-  return generateCstDts(parser.getGAstProductions(), {
+  const generatedFileWarning = `
+// This file is generated by chevrotain. Do not edit it directly.
+// To change syntax tree definitions, edit sii-parser.ts,
+// then execute \`npm run gen:types -w packages/clis/parser\`.
+
+`;
+  return generatedFileWarning + generateCstDts(parser.getGAstProductions(), {
```

packages/libs/map/types.ts
nautofon: Very cool, thank you!

packages/libs/map/types.ts
nautofon: The usual axes of a spatial cartesian coordinate system are *easting-northing-up*. Since SCS's choice of *easting-up-southing* is quite uncommon, using clear names for coordinate properties and variables is particularly important. In my view, `x` and `y` don't live up to that.

Even so, consistency is a winning argument, I suppose.

Renaming, and adding a few comments.

packages/clis/parser/game-files/sii-visitors.ts
nautofon: I wrote down your preference in my notes and will try to remember it next time.

In this particular case though, I think it has to be `null` because
- `null` and `undefined` are distinct concepts in JavaScript, and
- the schema validation works off of JSON types, which don't include `undefined`.

packages/clis/parser/game-files/sii-visitors.ts
nautofon: Ahyes.

Incidentally, the reason why I got confused about the order is that the order of token definitions at the top of this file doesn't match that of `objectPropertyValue`.

https://github.com/truckermudgeon/maps/blob/885e0cf9992cfb54da95f0ecfe0f011762480299/packages/clis/parser/game-files/sii-parser.ts#L27-L41

packages/clis/parser/game-files/sii-parser.ts
nautofon: Changing the order.

packages/clis/parser/game-files/def-parser.ts
nautofon: My rule of thumb for `debug` vs. `info` or higher is: "Does a regular user want to know this?"

In this case, position information simply doesn't exist in the sii file. I suspect none of these mileage targets are in fact used by the game. Even if they were, there's literally nothing that can be done about it. I think `debug` is exactly the right logging level for this one.

packages/clis/parser/game-files/map-files-parser.ts
nautofon: Maybe. Does a regular user want to know this? I can't really say.

The argument **for** would be that some users might be able to fix these missing positions on their end. For example, a few of the missing node UIDs did exist in earlier game versions. If you have a list of node UIDs and their positions from like 1.49, you can use it to fix some of these targets. (Though I suspect there aren't many users who would be able to do this.)

The argument **against** is that it clutters up the log. Users who don't have every DLC are particularly affected. If you run this without `includeDlc`, you'll get like 700 lines of "node not found" entries that may look scary, but are ultimately meaningless if you don't own the DLC anyway.

I'd tend towards keeping it as `debug`.

By the way, I found it easy to set the logging level with an environment variable. I don't think a `parser` option is necessary, but perhaps this might be mentioned in the readme.
```sh
CONSOLA_LEVEL=4 npx parser ...
```

packages/clis/parser/game-files/map-files-parser.ts
nautofon: Ahyes, well spotted.

I wrote it like this intentionally because my original goal was to be able to re-create [mileage-targets.csv](https://github.com/nautofon/ats-towns/blob/california/mileage-targets/mileage-targets.csv). But you're quite right: If we already have coordinates, `nodeUid` is entirely unnecessary, so it shouldn't be in the output file.

packages/clis/parser/game-files/sii-visitors.ts
nautofon: Ahyes; adding return type to the signature.

packages/clis/parser/game-files/def-parser.ts
nautofon: Ahyes.

packages/clis/parser/game-files/sii-schemas.ts
nautofon: Ahyes.

I ended up with `[number, number, number] | [null, null, null]`, which seems more appropriate for a JSON schema.

nautofon:
Thank you so much for your detailed comments!

I've applied most of your suggested changes. Happy to talk more about the changes that I haven't applied; see above.

Would you like me to squash some of these commits before merging, once we get that far?

packages/clis/parser/types/sii-visitor.d.ts
truckermudgeon: Ahhhh... I remember thinking I should do this shortly after committing https://github.com/truckermudgeon/maps/commit/492b2fe5aa268090923ef2c09274c6060e9864b2, then quickly forgot about it. Thank you for bringing this up 🙂 

Done in 520a0d6.

packages/clis/parser/game-files/sii-visitors.ts
truckermudgeon: Both fair points 👍. Leaving this as `null` SGTM.

packages/clis/parser/game-files/map-files-parser.ts
truckermudgeon: > By the way, I found it easy to set the logging level with an environment variable. I don't think a `parser` option is necessary, but perhaps this might be mentioned in the readme.

🤯 I did not know about the `CONSOLA_LEVEL` env var... and is why I've been overly-favoring warnings; I am almost always in "debug" mode, myself.

I think you're absolutely right when you say:

> My rule of thumb for debug vs. info or higher is: "Does a regular user want to know this?"

and I will re-audit the `logger.warn`s in use 😅

packages/clis/parser/game-files/sii-visitors.ts
truckermudgeon: Ahhh... in that case, could you also re-arrange the order of token definitions to match that of `objectPropertyValue`?

truckermudgeon (approved):
LGTM!

Thanks for taking the time to discuss the less-trivial comments. I have learned a couple of things 🙂.

> Would you like me to squash some of these commits before merging, once we get that far?

My preference for feature PRs is to squash everything down to a single commit:
* if any more context about the individual commits is needed, then the PR page is a click away with those `(#31)`-style links GitHub adds to the commit message on merge
* I am not disciplined enough to always write high-quality commit messages when working on a new feature, or to make sure my commits address only one "topic" of changes (and so squashing hides my mess 🙃)

But I think the commits and messages you have now are great, in both grouping and wording, and could stand alone as separate commits to `main`.

The repo _did_ enforce my squash-only preference. But I've just loosened that setting, and will leave it up to you to pick 🙂

packages/clis/parser/game-files/sii-visitors.ts
truckermudgeon: Actually, nevermind... you won't be able to do that because of declaration issues:

image

Ignore my previous comment, and pay no attention to the ~man behind the curtain~ order of token definitions 😅.

packages/clis/parser/game-files/map-files-parser.ts
nautofon: I mean, you *are* the primary user of this software, right? I think it's fair to optimize for that. A re-audit isn't super urgent in my view.

For what it's worth, I found the env var [somewhere deep down in the Consola readme](https://github.com/unjs/consola/blob/faa9cbd29d0fe89c5919d20c81c64660a04eb6ac/README.md#log-level).

nautofon:
Well, if it was me, I guess I'd just rebase onto main, leaving the individual commits as-is. But it's your project, so it's your choice to make.

Thanks for explaining your thought process though. I tend to rebase dev branches a lot while working, and the commit hooks on this repository made that a bit more cumbersome than I'm otherwise used to. Now that I know you usually like to squash anyway, I can make my life easier next time by following your pattern.

Also, thank you for your kind and encouraging comments throughout this thread! ☺️

truckermudgeon:
> it's your project, so it's your choice to make.

In that case, I'll go ahead and squash 😅

> the commit hooks on this repository made that a bit more cumbersome than I'm otherwise used to

Ha, I've been caught with that, too... I end up stashing changes that don't lint. I'll likely end up removing those hooks, nwo that I've set up GitHub actions and branch restrictions.

> thank you for your kind and encouraging comments throughout this thread! ☺️

Thank you for the level of quality in your contribution and discussion! It reminded me of PRs at my day job raised by staff engineers ☺️

nautofon:
Just one more thing: I have another branch, [`mileage-targets-geojson`](https://github.com/nautofon/maps-nodejs/tree/mileage-targets-geojson), that uses the intermediate array file to create a GeoJSON file you can load into QGIS.

Now, I wrote this primarily to see how *using* the parsed data would feel like. A proof-of-concept, basically. I doubt this would be a good fit for merging into `main`. I might just throw it away.

But if you *wanted* to have this functionality in your software, I could clean it up and open another PR for it. In that case, feel free to let me know.

truckermudgeon:
> A proof-of-concept, basically. I doubt this would be a good fit for merging into `main`.

👀  thanks for letting me know about that branch  👀 

I'll say, "no, thanks" for now, to limit the set of `generator` outputs to ones that are actively being consumed by other projects in the repo.

PR #34, opened 2024-12-28
generator: add routing support for ferries

truckermudgeon:
This PR updates the `generator graph` command to include information about ferry terminals.

Similar to a company POI that isn't located at the end of a road, a ferry terminal is added to the routing graph as a node with edges to:
* the other ferry terminal(s) to which the ferry connect, and
* the nearest route-able road node

A future PR will update the demo app to use this data, and visualize routes that require taking a ferry.

### DIFF
--- a/packages/clis/generator/commands/graph.ts
+++ b/packages/clis/generator/commands/graph.ts
@@ -66,9 +66,9 @@ export async function handler(args: BuilderArguments<typeof builder>) {
     includeHidden: false,
   });
 
-  const graph = generateGraph(tsMapData);
+  const graph = generateGraph(tsMapData, args.map);
   if (args.check) {
-    await checkGraph(graph, tsMapData);
+    await checkGraph(graph, tsMapData, args.map);
   }
 
   if (!args.dryRun) {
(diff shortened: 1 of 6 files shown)


packages/clis/generator/commands/graph.ts
```diff
@@ -66,9 +66,9 @@ export async function handler(args: BuilderArguments<typeof builder>) {
     includeHidden: false,
   });
 
-  const graph = generateGraph(tsMapData);
+  const graph = generateGraph(tsMapData, args.map);
```
truckermudgeon, line 69: 🤔 maybe `args.map` can be part of `tsMapData`. Then it wouldn't be possible to call, e.g., `generateGraph(atsMapData, 'europe')`.

packages/clis/generator/graph/check-graph.ts
```diff
@@ -25,12 +28,16 @@ interface Unrouteable {
...
+  const { nodes, companies, prefabs, cities } = tsMapData;
 
-  // check that all companies can be reached from some random company.
-  const allCompanies = [...companies.values()].filter(company =>
-    prefabs.has(company.prefabUid.toString(16)),
+  // check that all companies in known cities can be reached from some
+  // random company.
+  const allCompanies = [...companies.values()].filter(
+    company =>
+      cities.has(company.cityToken) &&
```
truckermudgeon, line 39: This extra filtering condition cleans up a lot of `generator graph` log spam about unreachable, future-DLC cities.

Issue #36, opened 2024-12-29
Possible inaccuracy in projection math


nautofon:
It's fantastic that you managed to figure out the meaning of `map_factor`! 

However, after taking a closer look at 360b8a8eafe07794231d1406a1f85523c59b1d9d ([map/projections.ts](https://github.com/truckermudgeon/maps/blob/360b8a8eafe07794231d1406a1f85523c59b1d9d/packages/libs/map/projections.ts)) today, I'm confused:

* `map_factor` has two components, suggesting a slightly different scale for the north/south axis vs. the east/west axis. But the calculation of `atsScale` only uses one of these components.

* `fromWgs84ToAtsConverter` uses formulas for the sphere, but `lengthOfDegreeAt` uses formulas for the ellipsoid. That doesn't look right.

* The name of `fromWgs84ToAtsConverter` suggests that it performs a datum transformation, but it actually doesn't.

I'm not sure any of this matters *too* much. The only important thing seems to be that our projects use more or less the same projection math, to ensure interoperability. Still, I have a general interest in coordinate reference systems, and I'd like to understand what SCS is doing here. If you made the choices above for particular reasons: Could you tell me what they are, please?

For what it's worth, I *suspect* that SCS uses a sphere, and that the scale *is* different for each axis. But I can't prove it.

I also don't know the sphere's radius. Both of our projects currently assume 6370.997 km, but the radius used by SCS might be a different one.

truckermudgeon:
> If you made the choices above for particular reasons: Could you tell me what they are, please?

The **TL;DR** for those choices is: they seemed to work/produce good-looking results 😅 

I must confess: I am definitely no expert in coordinate reference systems and only have a vague and probably-mistaken understanding of what terms like CRS, datum, and projection mean 😬. So apologies for any misuse of the terms below (corrections would be very welcome!):

I found myself using WGS 84 lat/lon coordinates because AFAIK, `maplibre` only speaks Web Mercator. So I muddled my way through figuring out how to convert game coordinates into something that looked like lat/lon coordinates by playing around with `proj4` and suffering through a lot of trial and error. 

Originally, the `+k_0` parameter in the `proj4` string I arrived at was based on generating GeoJSON for the road network, loading it up in QGIS, seeing how it aligned with a map of the US, tweaking the scale, and then repeating until I got something that lined up "ok".

[...]

PR #39, opened 2025-01-23
map: update lat/lon calculations

truckermudgeon:
See https://github.com/truckermudgeon/maps/issues/36 for more context.

### DIFF
--- a/packages/libs/map/projections.ts
+++ b/packages/libs/map/projections.ts
@@ -1,8 +1,8 @@
 import type { Position } from '@truckermudgeon/base/geom';
 import * as proj4 from 'proj4';
 
-// assumes `Normal Sphere` definition of Earth
-// https://proj.org/en/stable/usage/ellipsoids.html#built-in-ellipsoid-definitions
+// The Earth radius is canceled out in the coordinate calculation, so the exact
+// value doesn't matter. The chosen radius is the Clarke 1866 Authalic Sphere.
 const earthRadiusMeters = 6_370_997;
 const lengthOfDegree = (earthRadiusMeters * Math.PI) / 180;
 
--- a/packages/libs/map/projections.ts
+++ b/packages/libs/map/projections.ts
@@ -1,6 +1,10 @@
 import type { Position } from '@truckermudgeon/base/geom';
-import { toRadians } from '@truckermudgeon/base/geom';
 import * as proj4 from 'proj4';
+
+// assumes `Normal Sphere` definition of Earth
+// https://proj.org/en/stable/usage/ellipsoids.html#built-in-ellipsoid-definitions
+const earthRadiusMeters = 6_370_997;
+const lengthOfDegree = (earthRadiusMeters * Math.PI) / 180;
 
 // from def/climate.sii
 const atsDefData = {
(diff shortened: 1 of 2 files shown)


packages/libs/map/projections.ts
```diff
@@ -10,24 +14,21 @@ const atsDefData = {
...
-  `+k_0=${1 / atsScale}`,
 ].join(' ');
 
 const fromWgs84ToAtsConverter = proj4.default(ats);
 export const fromAtsCoordsToWgs84 = ([x, y]: Position): Position => {
-  // ATS coords are like LCC coords, except in ATS coords Y grows southward (its sign is reversed).
-  const lccCoords: Position = [x, -y];
+  const lccCoords: Position = [
+    x * atsDefData.mapFactor[1] * lengthOfDegree, // ~19.647 ┬ avg ~19.668
+    y * atsDefData.mapFactor[0] * lengthOfDegree, // ~19.688 ┘
```
truckermudgeon, line 30: It's nice not to have to worry about flipping signs and using `Math.abs` 🙂

packages/libs/map/projections.ts
```diff
@@ -40,56 +41,37 @@ const ets2DefData = {
...
+  // find any clues in the def files.
+  const calais = [-31100, -5500];
+  const isUk =
+    x * ukScaleFactor < calais[0] && //
+    y * ukScaleFactor < calais[1];
   if (isUk) {
-    x -= 16_650; // bigger offset => push UK stuff left
-    y -= 2_700; // smaller offset => push UK stuff down
+    x = (x + calais[0] / 2) * ukScaleFactor;
+    y = (y + calais[1] / 2) * ukScaleFactor;
```
truckermudgeon, line 69: In re-examining the numbers, I noticed these were so close to half of the Calais coords... and the map produced using the half-values looks good. So yay, fewer magic numbers 🎉 (now, it's just a magic calculation 🙃).

packages/libs/map/projections.ts
```diff
@@ -10,24 +14,21 @@ const atsDefData = {
...
   mapFactor: [-0.00017706234, 0.000176689948],
 } as const;
-const atsScale = Math.abs(
-  lengthOfDegreeAt(atsDefData.mapOrigin[0]) * atsDefData.mapFactor[0],
-); // 19.65665620539649
 const ats = [
   '+proj=lcc', // lambert conformal conic
-  '+units=m',
-  '+ellps=sphere',
+  `+R=${earthRadiusMeters}`,
```
truckermudgeon, line 19: I figured it'd be nice to make this proj string as parameterized as possible, so I went with the `+R` parameter over the hardcoded `+ellps=sphere` parameter.

packages/libs/map/projections.ts
```diff
@@ -40,56 +41,37 @@ const ets2DefData = {
...
-  // N.B.: all the UK detection and offsetting are just guesses.
-  // Couldn't find anything in def files to do this in a more accurate way.
-
-  // treat all coords up-and-to-the-left of Calais as UK coords
-  const calais = [-31100, -5500];
-  const isUk = x < calais[0] && y < calais[1];
-  const converter = isUk ? fromWgs84ToUkConverter : fromWgs84ToEts2Converter;
   // apply mapOffset to coords before projecting.
   x -= ets2DefData.mapOffset[0];
   y -= ets2DefData.mapOffset[1];
```
truckermudgeon, line 56: I am sad that `def/climate.sii`'s `mapOrigin` and `mapFactor` are `[y, x]` but `mapOffset` is `[x, y]`. The inconsistency irks me.

truckermudgeon:
Hey @nautofon! Would you be able to take a look at this to double-check my math / comments?

packages/libs/map/projections.ts
nautofon: Semantically, ”add half of Calais” doesn’t make sense to me at all. I might have preferred the old way, where the magic at least was more obvious.

But for all I know, this might be *exactly* how SCS implemented it…

For what it’s worth, I want to try a few things with the ETS2 demo version, but haven’t gotten around to it. Not sure if I can learn something new there, but I’ll let you know if I do.

packages/libs/map/projections.ts
nautofon: `sphere` is kinda magical. I think in this case, specifying the radius directly is the right call.

packages/libs/map/projections.ts
nautofon: Same. 🤨

packages/libs/map/projections.ts
```diff
@@ -1,7 +1,11 @@
 import type { Position } from '@truckermudgeon/base/geom';
-import { toRadians } from '@truckermudgeon/base/geom';
 import * as proj4 from 'proj4';
 
+// assumes `Normal Sphere` definition of Earth
+// https://proj.org/en/stable/usage/ellipsoids.html#built-in-ellipsoid-definitions
+const earthRadiusMeters = 6_370_997;
```
nautofon: May I suggest changing the comment to the following:
```js
// The Earth radius is canceled out in the coordinate calculation, so the exact
// value doesn't matter. The chosen radius is the Clarke 1866 Authalic Sphere.
```

If you wanted to, you could choose another radius, for example that of the [GRS 1980 Authalic Sphere](https://epsg.org/ellipsoid_7048/GRS-1980-Authalic-Sphere.html), which is 6371007 meters. But since the term is canceled, it really doesn’t matter.

nautofon (approved):
Looks good to me overall. Thanks for doing this!

packages/libs/map/projections.ts
truckermudgeon: > ”add half of Calais” doesn’t make sense to me at all. I might have preferred the old way, where the magic at least was more obvious.

That's a fair call-out... I'll leave it as-is for now, because I _think_ the calculation has something to do with "half the distance from 0,0 to Calais in game coordinates"... like, the transform that's happening is a translate toward the halfway point between Calais and 0,0, and then a scale about 0,0. Or a scale about 0,0, and then a translate... I can't maths this late 😅.

> For what it’s worth, I want to try a few things with the ETS2 demo version, but haven’t gotten around to it. Not sure if I can learn something new there, but I’ll let you know if I do.

Please do. I may play around with the scale-and-translate / translate-and-scale idea later... I'll let you know if I come up with something, too.

PR #40, opened 2025-01-24
navigator: add wip components

truckermudgeon:
This PR adds the components I've been building for the `navigator` project. It's all still very much a work-in-progress: props are likely to change, not all components have dark-mode styling, etc.

The Storybook can be viewed by running:

```sh
# from the repo root:
npm run storybook -w packages/apps/navigator
```

Screenshot 2025-01-23 at 9 04 27 PM

PR #42, opened 2025-01-25
navigator: add app, fake mode

truckermudgeon:
This PR wires up the components added in https://github.com/truckermudgeon/maps/pull/40 into a minimally functioning web app. 

The `navigator` project can be run in either:

#### Real mode

Real mode expects the `telemetry` and `navigation` backends re-introduced in #41 to be running. Only player tracking is supported at the moment.

image

#### Fake mode

Similar to the `navigator` project's [Storybook](https://github.com/truckermudgeon/maps/pull/40), fake mode is meant to help test / build out the UI. Fake mode doesn't require the `telemetry` or `navigation` backends to be running, and is accessed with the `?fake` query parameter:

Screenshot 2025-01-25 at 9 30 43 PM

### DIFF
--- a/packages/apps/navigator/src/controllers/app.ts
+++ b/packages/apps/navigator/src/controllers/app.ts
@@ -134,7 +134,6 @@ export class AppControllerImpl implements AppController {
         prevBearing = currBearing;
         currBearing = bearing;
 
-        // TODO do this in a reaction / observable context?
         switch (store.cameraMode) {
           case CameraMode.FOLLOW:
             map.easeTo({
--- a/packages/apps/navigator/src/controllers/app.ts
+++ b/packages/apps/navigator/src/controllers/app.ts
@@ -2,7 +2,6 @@ import { assertExists } from '@truckermudgeon/base/assert';
 import type { Position } from '@truckermudgeon/base/geom';
 import { getExtent } from '@truckermudgeon/base/geom';
 import { Preconditions, UnreachableError } from '@truckermudgeon/base/precon';
-import { fromAtsCoordsToWgs84 } from '@truckermudgeon/map/projections';
 import type { Route, RouteDirection } from '@truckermudgeon/navigation/types';
 import type { Marker } from 'maplibre-gl';
 import { action, makeAutoObservable, observable } from 'mobx';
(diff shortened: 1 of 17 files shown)


packages/apps/navigator/src/controllers/app.ts
```diff
@@ -0,0 +1,304 @@
...
+        const { speedMph, position: center, bearing } = gameState;
+        if (prevPosition.every(v => !v)) {
+          map.setCenter(center);
+        }
+        prevPosition = currPosition;
+        currPosition = center;
+        prevBearing = currBearing;
+        currBearing = bearing;
+
+        // TODO do this in a reaction / observable context?
```
truckermudgeon:
```suggestion
```

I tried to make player pos an observable field on a store, and then have the PlayerMarker component act as an observer on that field.

It didn't really work out... the Marker didn't move smoothly 😕

packages/apps/navigator/src/controllers/controls.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  get showSearchFab(): boolean {
+    return !!this.appStore.activeRoute && !this.appStore.showNavSheet;
+  }
+}
+
+export class ControlsControllerImpl implements ControlsController {
+  startListening(store: ControlsStore, appClient: AppClient) {
+    // TODO keep returned Unsubscribable
```
truckermudgeon:
```suggestion
```

Not really necessary... this controller is a singleton in practice, and is never disposed during the lifetime of the Navigator app.

packages/apps/navigator/src/controllers/nav-sheet.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+export class NavSheetControllerImpl implements NavSheetController {
+  constructor(private readonly appClient: AppClient) {}
+
+  onBackClick(store: NavSheetStore) {
+    switch (store.currentPageKey) {
+      case NavPageKey.CATEGORIES:
+        // shouldn't be able to press 'Back' from the 'types' page.
```
truckermudgeon:
```suggestion
        // shouldn't be able to press 'Back' from the 'Categories' page.
```

PR #43, opened 2025-02-01
generator,maps: deserialize uids as BigInts

truckermudgeon:
When producing JSON files, `parser` serializes bigint UIDs as [hex strings](https://github.com/truckermudgeon/maps/blob/cc33a70f7fdd633e574c36256e32995d34c03ad2/packages/clis/parser/index.ts#L104-L106).

But when reading these JSON files, `generator` deserializes these hex strings as strings, and not as bigints, which introduces inconsistency/incorrectness. For example, working with a `Node` object deserialized from `usa-nodes.json`:
- the [types](https://github.com/truckermudgeon/maps/blob/cc33a70f7fdd633e574c36256e32995d34c03ad2/packages/libs/map/types.ts#L14) _say_ that `node.uid` is a bigint, even though the in-memory object has it stored as a string
- the map of Node objects produced by `readMapData` uses string keys, so to query the map, one needs to do something like:
```ts
const someNode: Node = getSomeNodeByReadingJsonFile();
const isKnownNode = tsMapData.nodes.has(someNode.uid.toString(16));
```
which is annoying, because `someNode.uid` is _already_ a string, so there's really no need to call `.toString(16)` on it. The `.toString(16)` is needed only to satisfy the compiler.

This PR addresses this inconsistency/incorrectness by updating the JSON-reading code in `readArrayFile` to deserialize UIDs as bigints. Note that it relies on a few hard-coded [assumptions](https://github.com/truckermudgeon/maps/blob/cc33a70f7fdd633e574c36256e32995d34c03ad2/packages/clis/generator/read-array-file.ts#L44-L57):
- a field named `uid` or whose name ends in `Uid` contains a hex-encoded bigint string
- a field whose name ends in `Uids` contains an array of hex-encoded bigint strings

The end result is the removal of almost all `someUid.toString(16)` calls (the only ones remaining are for debug logging, and for GeoJSON generation).

### DIFF
--- a/packages/clis/generator/mapped-data.ts
+++ b/packages/clis/generator/mapped-data.ts
@@ -67,34 +67,6 @@ type PickKey<
   U extends keyof MapData[T][0],
 > = MapData[T][0][U];
 
-//const mapDataKeyProps = {
-//  cities: 'token',
-//  companies: 'uid',
-//  companyDefs: 'token',
-//  countries: 'token',
-//  cutscenes: 'uid',
-//  dividers: 'uid',
-//  ferries: 'token',
-//  mapAreas: 'uid',
-//  mileageTargets: 'token',
-//  modelDescriptions: 'token',
-//  models: 'uid',
-//  nodes: 'uid',
-//  prefabDescriptions: 'token',
-//  prefabs: 'uid',
-//  roadLooks: 'token',
-//  roads: 'uid',
-//  routes: 'token',
-//  trajectories: 'uid',
-//  triggers: 'uid',
-//  achievements: 'token',
-//  pois: undefined,
-//  elevation: undefined,
-//} satisfies Record<keyof MapData, 'token' | 'uid' | undefined>;
-//
-//type KeyTypeFor<T extends keyof MapData> =
-//  MapData[T][0][(typeof mapDataKeyProps)[T]];
-
 interface MapDataKeyFields {
   achievements: PickKey<'achievements', 'token'>;
   cities: PickKey<'cities', 'token'>;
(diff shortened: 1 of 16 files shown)


packages/clis/generator/mapped-data.ts
```diff
@@ -62,19 +62,50 @@ const MapDataKeys: Record<keyof MapData, void> = {
 };
 const mapJsonFiles = Object.freeze(Object.keys(MapDataKeys));
 
+type PickKey<
+  T extends keyof MapData,
+  U extends keyof MapData[T][0],
+> = MapData[T][0][U];
+
+interface MapDataKeyFields {
```
truckermudgeon, line 70: I wanted to declare either a type or object that looked like:

```ts
{
  achievements: 'token',
  cities: 'token',
  companies: 'uid';
  // ...
}
```

but couldn't get the types to work... so had to settle for something that repeats the keys 😬

packages/clis/generator/read-array-file.ts
```diff
@@ -1,12 +1,57 @@
...
 ): T[] {
-  const results: unknown = JSON.parse(fs.readFileSync(path, 'utf-8'));
+  const basename = path.basename(filepath, '.json');
+  const start = Date.now();
+  const reviver =
+    basename.endsWith('elevation') ||
+    basename.endsWith('prefabDescriptions') ||
+    basename.endsWith('nodes')
+      ? undefined
+      : bigintReviver;
```
truckermudgeon, line 23: Custom revivers can add a lot of time to the deserialization process (~10 seconds on my laptop). Much of that penalty can be reduced by skipping really large JSON files:

- `elevation` is just an array of an array of numbers, so no point using a custom reviver
- `prefabDescriptions` is an array of deep/complex objects, but with no UID properties
- `nodes` is a _huge_ array of shallow objects... it's faster to special-case this file than to use a custom reviver (see lines 32--39)

Issue #44, opened 2025-02-03
Search bar doesn't work in the demo app when run locally for single game


nautofon:
There are three somewhat related issues here:

### 1. Generator `ets2-villages` not mentioned in readme
After generating the data files for the demo app exactly as described in the readme, there is no `ets2-villages.geojson`. The "cities" menu opened by the search bar has no selectable entries; it just says "No options". The demo app complains:
```
could not load cities/towns geojson.
```

The command `npx generator ets2-villages -o dirToWriteFileTo` should probably be added to the ["generator"](https://github.com/truckermudgeon/maps/blob/6632a4120c3b0da29245ff8729c914503d2e6916/README.md#generator) section in the readme.

### 2. Search bar depends on ETS2 city data being available
Even after adding `ets2-villages.geojson` to the app directory, the search bar still doesn't work. The demo app complains:
```
Uncaught (in promise) Error:
    assertExists assert.ts:3
    createSortedCityOptions CitySearchBar.tsx:177
...
```
That particular line expects to find country data for every town. But my `cities.geojson` file only contains ATS cities. I can't generate the combined file because I don't have ETS2.

https://github.com/truckermudgeon/maps/blob/6632a4120c3b0da29245ff8729c914503d2e6916/packages/apps/demo/src/CitySearchBar.tsx#L177

### 3. Achievement data files not included
Achievement search doesn't work out of the box, either. The demo app complains:
```
could not load achievements json.
```

On [truckermudgeon.github.io](https://github.com/truckermudgeon/truckermudgeon.github.io/tree/47187c15527a50e93d6d53bef131ed768cf27cdc), I see those files contain instructions for how to generate them. But that doesn't help much if the files are missing to begin with. Those instructions should probably be in the readme, or at least in `AchievementSearchBar.tsx`.

For the readme, perhaps it's sufficient to point out that these files are not included in this repository, and that they can be downloaded from truckermudgeon.github.io.

PR #45, opened 2025-02-03
Make it easier to set up the demo app offline with a working search bar

nautofon:
See issue #44.

### DIFF
--- a/packages/apps/demo/src/CitySearchBar.tsx
+++ b/packages/apps/demo/src/CitySearchBar.tsx
@@ -172,6 +172,7 @@ function createSortedCityOptions(
   }
 
   return cities
+    .filter(city => countryCodeToName.get(city.properties.countryCode))
     .map(({ properties, geometry }) => ({
       map: properties.map,
       state: assertExists(countryCodeToName.get(properties.countryCode)),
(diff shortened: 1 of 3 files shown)


packages/apps/demo/src/CitySearchBar.tsx
```diff
@@ -172,6 +172,7 @@ function createSortedCityOptions(
   }
 
   return cities
+    .filter(city => countryCodeToName.get(city.properties.countryCode))
```
truckermudgeon: Tiny nit: I have a slight preference for `.filter` predicates returning booleans, instead of values that can be coerced to booleans.
```suggestion
    .filter(city => countryCodeToName.has(city.properties.countryCode))
```

truckermudgeon (approved):
Thanks for acting as a guinea pig and testing out my directions 😅 (which assume so much about my personal setup).

packages/apps/demo/src/CitySearchBar.tsx
nautofon: Makes perfect sense. I appreciate you pointing this stuff out! The last time I did serious JavaScript was like 15 years ago. I'm still marveling at the `class` keyword, we used to do prototypes and worry about Internet Exploder 5.5 (cue grandpa story time 😅). Takes a while to get familiar with the modern language.

truckermudgeon: I, too, have worked with JS for a while 😂 . The language (and the ecosystem around the language) has indeed come a long way since the Internet Exploder years!

PR #47, opened 2025-02-09
Add map labels generator based on mileage targets

nautofon:
## Add module for generating map labels from mileage targets

My goal with this proposed change is two-fold:

1. Make `maps` able to generate labels automatically, on its own, without any external data sources.
2. Have an easy way to manually modify, remove, or add individual map labels.

The first part—automatic label generation—already works well enough that I think most users won't even notice the difference from the old data in [ats-towns](https://github.com/nautofon/ats-towns/tree/towns-txt).

Even so, the quality is not entirely satisfactory: Using mileage targets as a source inevitably means that some labels that I'd like to see on the map will be missing, and a number of those labels that *are* generated will be wrong in some way. That's why I think the ability to modify the labels manually is important. The module added in this change allows that by optionally reading a manually edited [metadata](https://github.com/nautofon/ats-towns/pull/6) table, which would contain such modifications.

I'm opening this draft PR now to give you a rough idea of what I'm even talking about. 🙂

But before we consider a merge or review, there is one big open question I think we should try to answer first:

### How would we organize the exchange of data?

* In principle, we could stick to what we do now: I provide an external GeoJSON file, which is directly used by the map. However, that would mean that I'd actually be the only user of this new module in your project. Even somebody who does a `git checkout` and runs the demo locally would still use "my" file over the internet, unless they fiddle with the URL in [SceneryTownSource.tsx](https://github.com/truckermudgeon/maps/blob/8c357a298eeb04d9f63cbd63c8bae67c7cdabd3d/packages/libs/ui/SceneryTownSource.tsx). Is that what we want?

    If so, I suppose it'd still be worth it to merge it in. If nothing else, having this module readily available in your codebase improves the [bus factor](https://en.wikipedia.org/wiki/Bus_factor), which will help if/when I stop updating that external file for whatever reason.

* One out of several other ways to organize this would be to include the metadata table in this repository, with me opening a new PR every time I think it should be updated. In this scenario, the GeoJSON file would be generated as part of the live deployment, probably in the Makefile.

    The primary disadvantage here is that any changes I make aren't visible on the live server until you deploy them. Plus, a large number of PRs seems tedious for the both of us. So I'm not a fan, but it's an option.

* Perhaps we want to chart a middle course. For example, [SceneryTownSource.tsx](https://github.com/truckermudgeon/maps/blob/8c357a298eeb04d9f63cbd63c8bae67c7cdabd3d/packages/libs/ui/SceneryTownSource.tsx) might continue to refer to the external file, but we keep a copy of the metadata table stored in this repository (thus making it potentially self-contained—bus factor). The copy wouldn't need to be updated on *every* metadata change; once or twice a year might be enough.

    But this approach would mean that this copy would be a file that exists in this repository, without actually being used for anything. It would just be a backup of sorts. Not sure if *that* idea convinces me, either.

What do you think?

### DIFF
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -63,14 +63,6 @@ export class LabelProducer {
   }
 
   /**
-   * Like {@link makeLabels}, but additionally filters out all invalid labels.
-   *
-   * @returns All _valid_ map labels for the provided game data and metadata.
-   */
-  makeValidLabels(): Label[] {
-  }
-
-  /**
    * Reads game data from the parser output into memory.
    * Suitable for feeding into the {@link LabelProducer} constructor.
    *
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -384,15 +384,17 @@ export class LabelDataProvider {
   }
 
   /**
-   * Search for a country in the game data by ISO 3166-1 or ISO 3166-2 code.
+   * Check whether a metadata record has an ISO 3166-1 or ISO 3166-2 code
+   * that matches a country in the provided game data.
    *
-   * @param isoCode - The country or country and subdivision code to look up.
+   * @param meta - The metadata record for which to look up the country /
+   *               country and subdivision ISO code.
    *
    * @returns True if the given code identifies a country in the game data.
    *
    * @see {@link clis/generator/geo-json/populated-places!ets2IsoA2}
    */
-  isValidCountry(isoCode: string): boolean {
+  hasKnownCountryCode(meta: LabelMeta): boolean {
   }
 
   /**
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -4,7 +4,12 @@
  * @author nautofon
  */
 
-import type { City, Country, MileageTarget } from '@truckermudgeon/map/types';
+import type {
+  City,
+  Country,
+  LabelMeta,
+  MileageTarget,
+} from '@truckermudgeon/map/types';
 import type { GeoJSON } from 'geojson';
 import type { MappedDataForKeys } from '../mapped-data';
 
@@ -92,124 +97,6 @@ export class LabelProducer {
 }
 
 /**
- * Metadata attributes for a map label.
- *
- * All attributes are optional. When applying metadata to labels generated
- * from mileage targets, an __undefined__ attribute (`null`) should cause
- * that attribute to be undefined in the result as well, whereas a
- * __missing__ attribute should cause the result to use the mileage target
- * data for that particular attribute.
- *
- * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
- */
-export interface LabelMeta {
-  // Meant for JSON, thus this interface must use null rather than undefined.
-
-  /**
-   * The token identifying the mileage target to apply the label attributes to.
-   *
-   * If missing or undefined, this object describes a new label instead.
-   */
-  token?: string | null;
-
-  /**
-   * The label text / feature name.
-   */
-  text?: string | null;
-
-  /**
-   * The adjusted easting, if any.
-   *
-   * Label metadata attributes use the terms easting and {@link southing} to
-   * refer to `x` / `y` coordinates. These more verbose terms avoid ambiguity
-   * of the coordinates' axis order and orientation. In the software project
-   * "Web-based maps for ATS and ETS2", only this interface {@link LabelMeta}
-   * and its implementers use these terms, in order to match the data files.
-   *
-   * The attributes easting and southing may be missing in metadata if the
-   * position read from mileage target data is already adequate.
-   */
-  easting?: number | null;
-
-  /**
-   * The adjusted southing, if any.
-   *
-   * @see {@link easting}
-   */
-  southing?: number | null;
-
-  /**
-   * The kind of location this label is for.
-   *
-   * Possible values include `city`, `town`, `unnamed`, and several others.
-   * Missing for most labels generated from new unassessed mileage targets;
-   * for such labels, the best value to assume as default is probably `town`.
-   *
-   * Label objects of the kind `unnamed` are not suitable for map display.
-   */
-  kind?: string | null;
-
-  /**
-   * Describes how the name is signed at a location in the game.
-   *
-   * Possible values are:
-   * - `all`:    Name well visible, no matter which direction you arrive from.
-   * - `most`:   Name visible when arriving from a clear majority of directions.
-   * - `some`:   Name visible in _some_ way, but it may not be very obvious.
-   * - `remote`: Name _not_ visible on site, but it appears on distance or
-   *             direction signs elsewhere.
-   */
-  signed?: 'all' | 'most' | 'some' | 'remote' | null;
-
-  /**
-   * True if a core part of the named location is accessible during regular
-   * gameplay.
-   */
-  access?: boolean | null;
-
-  /**
-   * True if the label is for a game location with deliverable industry, for
-   * example a scenery town with company depots (sometimes called a "suburb").
-   */
-  industry?: boolean | null;
-
-  /**
-   * The SCS token of the marked city this label can be proven to be associated
-   * with, if any.
-   */
-  city?: string | null;
-
-  /**
-   * The ISO 3166 code of the country / state / province the labeled feature
-   * is located in, for example `CZ` (Czechia) or `US-NV` (Nevada).
-   */
-  country?: string | null;
-
-  /**
-   * True (or missing) if it's recommended to show this label on the map
-   * by default. The value of this attribute is largely subjective.
-   */
-  show?: boolean | null;
-
-  /**
-   * The ISO 8601 date of the last time this location was checked in the game
-   * (usually `YYYY-MM`).
-   */
-  checked?: string | null;
-
-  /**
-   * Note or comment about the label or its attributes.
-   */
-  remark?: string | null;
-
-  /**
-   * Reference to real-life information about the labeled entity.
-   * Not currently used; reserved for future expansion.
-   */
-  ref?: unknown;
-}
-
-/**
  * Map label.
  *
  * In addition to the metadata attributes (which all are optional), this
@@ -265,7 +152,6 @@ export class GenericLabel implements Label {
   show: boolean | undefined;
   checked: string | undefined;
   remark: string | undefined;
-  ref: unknown;
 
   /**
    * @param data - The game data provider for the label's region.
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -153,13 +153,13 @@ export abstract class TargetLabel extends GenericLabel {
    * The original {@link MileageTarget} that this label was generated from.
    * @internal
    */
-  readonly target;
+  readonly target: MileageTarget;
 
   /**
    * Details on the mileage target analysis results.
    * @internal
    */
-  readonly analysis;
+  readonly analysis: TargetAnalysis;
 
   /**
    * @param target - The {@link MileageTarget} to generate a label for.
@@ -274,7 +274,7 @@ export class LabelDataProvider {
   /**
    * The full list of {@link MileageTarget} objects read from the parser output.
    */
-  readonly mileageTargets;
+  readonly mileageTargets: readonly MileageTarget[];
 
   /**
    * Function to transform game coordinates to geographic coordinates.
@@ -286,7 +286,11 @@ export class LabelDataProvider {
    *
    * @see {@link clis/generator/geo-json/normalize!createNormalizeFeature}
    */
-  readonly normalizeFeature;
+  readonly normalizeFeature: <
+    T extends GeoJSON.Feature<GeoJSON.Point, LabelMeta>,
+  >(
+    feature: T,
+  ) => T;
 
 
   /**
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -41,8 +41,7 @@ export class LabelProducer {
    *     The metadata records to use for augmenting the labels generated from
    *     mileage targets in the game data. Optional.
    *
-   * @see {@link clis/generator/mapped-data!readMapData}
-   * @see {@link readMapData}
+   * @see {@link LabelProducer.readMapData}
    * @see {@link readMetas}
    */
   constructor(
@@ -70,6 +69,8 @@ export class LabelProducer {
    * @param region - `'europe' | 'usa'`
    *
    * @returns The game map data, restricted to just the properties needed here.
+   *
+   * @see {@link clis/generator/mapped-data!readMapData}
    */
   static readMapData(
     dir: string,
@@ -83,6 +84,8 @@ export class LabelProducer {
    * @param jsonPath - The path of the metadata file. Only JSON is implemented.
    *
    * @returns An array of metadata records.
+   *
+   * @see {@link LabelDataProvider.readMetas}
    */
   static readMetas(jsonPath: string): LabelMeta[] {
   }
@@ -392,6 +395,11 @@ export interface TargetAnalysis {
  * This class gives access to parser results and to the metadata table.
  * It offers utility methods for easily querying particular aspects
  * needed for mileage target analysis.
+ *
+ * Metadata records with `country` attributes that don't match the
+ * game region of the loaded parser results are ignored. As a result,
+ * - one metadata file can apply to multiple regions, and
+ * - multiple metadata files can apply to a single region.
  */
 export class LabelDataProvider {
   // This class is somewhat inefficient, with nested sequential searches.
@@ -420,11 +428,9 @@ export class LabelDataProvider {
    *     The game map data to use as a primary source.
    * @param metas
    *     The metadata records to use for augmenting the labels generated from
-   *     mileage targets in the game data. Records with `country` attributes
-   *     that don't match the `gameData`'s region are ignored.
+   *     mileage targets in the game data.
    *
    * @see {@link clis/generator/mapped-data!readMapData}
-   * @see {@link readMapData}
    * @see {@link readMetas}
    */
   constructor(
@@ -541,21 +547,6 @@ export class LabelDataProvider {
    * @returns `'europe' | 'usa'`
    */
   region(): RegionName {
-  }
-
-  /**
-   * Reads game data from the parser output into memory.
-   * Suitable for feeding into the {@link LabelDataProvider} constructor.
-   *
-   * @param dir    - The path of the dir containing the parser output.
-   * @param region - `'europe' | 'usa'`
-   *
-   * @returns The game map data, restricted to just the properties needed here.
-   */
-  static readMapData(
-    dir: string,
-    region: RegionName,
-  ): MappedDataForKeys<['cities', 'countries', 'mileageTargets']> {
   }
 
   /**
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -27,7 +27,7 @@ type RegionName = 'usa' | 'europe';
  * );
  * const labels   = producer.makeLabels();
  * const features = labels
- *   .filter( label => label.isValid() )
+ *   .filter( label => label.isValid )
  *   .map( label => label.toGeoJsonFeature() );
  * ```
  *
@@ -106,7 +106,7 @@ export class LabelProducer {
  */
 export interface Label extends LabelMeta {
   /**
-   * @returns True if the map label is considered valid for display.
+   * True if the map label is considered valid for display.
    *
    * Examples for map labels that aren't valid:
    * - No coordinates are available.
@@ -115,7 +115,7 @@ export interface Label extends LabelMeta {
    *      for a location in a DLC that hasn't been released, or that it's
    *      located in a different game).
    */
-  isValid(): boolean;
+  readonly isValid: boolean;
 
   /**
    * @returns The metadata for the map label as a shallow copy of this object.
@@ -428,11 +428,10 @@ export class LabelDataProvider {
   }
 
   /**
-   * The name of the region for which game data is provided.
-   *
-   * @returns `'europe' | 'usa'`
+   * The name of the region for which game data is provided
+   * (`'europe'` or `'usa'`).
    */
-  region(): RegionName {
+  get region(): RegionName {
   }
 
   /**
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -322,19 +322,27 @@ export abstract class TargetLabel extends GenericLabel {
     if (analysis.country) {
       this.meta.country = analysis.countryCode;
     }
+
     if (analysis.city) {
       this.meta.kind = 'city';
       this.meta.city = analysis.city.token;
       this.meta.show = false;
     }
+
     if (
       analysis.excludeBorder ||
       analysis.excludeJunction ||
       analysis.excludeNumber
     ) {
       this.meta.kind = 'unnamed';
-      this.meta.text = this.target.editorName;
+
+      // Mention all mileage target name variants for excluded labels,
+      // to assist manual analysis.
+      const target = this.target;
+      this.meta.remark = [target.defaultName, ...target.nameVariants].join(';');
+      this.meta.text = target.editorName;
     }
+
     if (analysis.tooMuchDistance) {
       this.meta.access = false;
       this.meta.show = false;
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -113,7 +113,7 @@ export interface Label {
   readonly isValid: boolean;
 
   /**
-   * @returns The metadata for the map label as a shallow copy of this object.
+   * The metadata for the map label.
    *
    * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
    */
@@ -183,7 +183,8 @@ export class AtsLabel extends TargetLabel {
 /**
  * Map label generated from Euro Truck Simulator 2 mileage targets.
  *
- * Contains heuristics for ETS2 `editorName` practice and the ISO country code.
+ * Contains heuristics for ETS2 `editorName` practice, exclusion of "unnamed"
+ * country border targets, and the ISO country code.
  */
 export class Ets2Label extends TargetLabel {
 }
@@ -235,7 +236,8 @@ export interface TargetAnalysis {
   cityTokenEditorName?: string;
 
   /**
-   * True if the mileage target seems to be for an unnamed state line crossing.
+   * True if the mileage target seems to be for an unnamed state line
+   * or country border crossing.
    */
   excludeBorder?: boolean;
 
@@ -278,7 +280,6 @@ export class LabelDataProvider {
 
   /**
    * Function to transform game coordinates to geographic coordinates.
-   * Uses the precision given by {@link geographicCoordinatePrecision}.
    *
    * @param feature - A GeoJSON.Feature with (projected) game coordinates.
    *
@@ -371,7 +372,7 @@ export class LabelDataProvider {
    *
    * Country codes in mileage target tokens mostly follow ISO 3166
    * (but UK is used for Great Britain and XK for Kosovo).
-   * In the game data, ATS uses ISO subdivisions for country codes, while
+   * In the game's country data, codes used for ATS are ISO subdivisions, while
    * ETS2 uses the Distinguishing Sign of Vehicles in International Traffic.
    *
    * @param code - The mileage target country code to look up.
@@ -407,8 +408,6 @@ export class LabelDataProvider {
    * - `true` if the metadata record's country is in this game region.
    * - `false` if the metadata record's country is in another game region.
    * - `undefined` if the metadata record doesn't identify a country.
-   *
-   * @see {@link clis/generator/geo-json/populated-places!ets2IsoA2}
    */
   isInRegion(meta: LabelMeta): boolean | undefined {
   }
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -98,13 +98,8 @@ export class LabelProducer {
 
 /**
  * Map label.
- *
- * In addition to the metadata attributes (which all are optional), this
- * interface defines methods that all map labels should have available.
- *
- * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
  */
-export interface Label extends LabelMeta {
+export interface Label {
   /**
    * True if the map label is considered valid for display.
    *
@@ -119,8 +114,10 @@ export interface Label extends LabelMeta {
 
   /**
    * @returns The metadata for the map label as a shallow copy of this object.
+   *
+   * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
    */
-  meta(): LabelMeta;
+  readonly meta: LabelMeta;
 
   /**
    * @returns The map label as a GeoJSON point feature.
@@ -134,25 +131,8 @@ export interface Label extends LabelMeta {
 
 /**
  * Map label base class.
- *
- * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
  */
 export class GenericLabel implements Label {
-
-  token: string | undefined;
-  text: string | undefined;
-  easting: number | undefined;
-  southing: number | undefined;
-  kind: string | undefined;
-  signed: 'all' | 'most' | 'some' | 'remote' | undefined;
-  access: boolean | undefined;
-  industry: boolean | undefined;
-  city: string | undefined;
-  country: string | undefined;
-  show: boolean | undefined;
-  checked: string | undefined;
-  remark: string | undefined;
-
   /**
    * @param data - The game data provider for the label's region.
    */
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -122,7 +122,7 @@ export interface Label {
   /**
    * @returns The map label as a GeoJSON point feature.
    *
-   * @throws ReferenceError if no coordinates are available.
+   * @throws Error if no coordinates are available.
    *
    * @see {@link isValid}
    */
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -18,7 +18,7 @@ type RegionName = 'usa' | 'europe';
  * ```ts
  * const producer = new LabelProducer({
  *   gameData: LabelProducer.readMapData('path/to/parser-output', 'usa'),
- *   metaData: LabelProducer.readMetaData('path/to/meta.json'),
+ *   metas: LabelProducer.readMetas('path/to/meta.json'),
  * });
  * const labels   = producer.makeLabels();
  * const features = labels
@@ -37,20 +37,20 @@ export class LabelProducer {
   /**
    * @param __namedParameters.gameData
    *     The game map data to use as a primary source.
-   * @param __namedParameters.metaData
-   *     The metadata table to use for augmenting the labels generated from
+   * @param __namedParameters.metas
+   *     The metadata records to use for augmenting the labels generated from
    *     mileage targets in the game data. Optional.
    *
    * @see {@link clis/generator/mapped-data!readMapData}
    * @see {@link readMapData}
-   * @see {@link readMetaData}
+   * @see {@link readMetas}
    */
   constructor({
     gameData,
-    metaData = [],
+    metas = [],
   }: {
     gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>;
-    metaData?: LabelMeta[];
+    metas?: LabelMeta[];
   }) {
   }
 
@@ -89,13 +89,13 @@ export class LabelProducer {
   }
 
   /**
-   * Reads the metadata table from a file into memory.
+   * Reads metadata records from disk into memory.
    *
    * @param jsonPath - The path of the metadata file. Only JSON is implemented.
    *
-   * @returns The metadata table as array.
+   * @returns An array of metadata records.
    */
-  static readMetaData(jsonPath: string): LabelMeta[] {
+  static readMetas(jsonPath: string): LabelMeta[] {
   }
 }
 
@@ -429,18 +429,18 @@ export class LabelDataProvider {
   /**
    * @param gameData
    *     The game map data to use as a primary source.
-   * @param metaData
-   *     The metadata table to use for augmenting the labels generated from
+   * @param metas
+   *     The metadata records to use for augmenting the labels generated from
    *     mileage targets in the game data. Records with `country` attributes
    *     that don't match the `gameData`'s region are ignored.
    *
    * @see {@link clis/generator/mapped-data!readMapData}
    * @see {@link readMapData}
-   * @see {@link readMetaData}
+   * @see {@link readMetas}
    */
   constructor(
     gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>,
-    metaData: LabelMeta[],
+    metas: LabelMeta[],
   ) {
   }
 
@@ -570,12 +570,12 @@ export class LabelDataProvider {
   }
 
   /**
-   * Reads the metadata table from a file into memory.
+   * Reads metadata records from disk into memory.
    *
    * @param jsonPath - The path of the metadata file. Only JSON is implemented.
    *
-   * @returns The metadata table as array.
+   * @returns An array of metadata records.
    */
-  static readMetaData(jsonPath: string): LabelMeta[] {
+  static readMetas(jsonPath: string): LabelMeta[] {
   }
 }
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -16,10 +16,10 @@ type RegionName = 'usa' | 'europe';
  *
  * @example
  * ```ts
- * const producer = new LabelProducer({
- *   gameData: LabelProducer.readMapData('path/to/parser-output', 'usa'),
- *   metas: LabelProducer.readMetas('path/to/meta.json'),
- * });
+ * const producer = new LabelProducer(
+ *   LabelProducer.readMapData('path/to/parser-output', 'usa'),
+ *   LabelProducer.readMetas('path/to/meta.json'),
+ * );
  * const labels   = producer.makeLabels();
  * const features = labels
  *   .filter( label => label.isValid() )
@@ -35,9 +35,9 @@ export class LabelProducer {
   readonly dataProvider: LabelDataProvider;
 
   /**
-   * @param __namedParameters.gameData
+   * @param gameData
    *     The game map data to use as a primary source.
-   * @param __namedParameters.metas
+   * @param metas
    *     The metadata records to use for augmenting the labels generated from
    *     mileage targets in the game data. Optional.
    *
@@ -45,13 +45,10 @@ export class LabelProducer {
    * @see {@link readMapData}
    * @see {@link readMetas}
    */
-  constructor({
-    gameData,
-    metas = [],
-  }: {
-    gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>;
-    metas?: LabelMeta[];
-  }) {
+  constructor(
+    gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>,
+    metas?: LabelMeta[],
+  ) {
   }
 
   /**
--- a/packages/clis/generator/geo-json/extra-labels.ts
+++ b/packages/clis/generator/geo-json/extra-labels.ts
@@ -10,8 +10,13 @@ import type {
   LabelMeta,
   MileageTarget,
 } from '@truckermudgeon/map/types';
+import fs from 'fs';
 import type { GeoJSON } from 'geojson';
+import { logger } from '../logger';
 import type { MappedDataForKeys } from '../mapped-data';
+import { readMapData } from '../mapped-data';
+import { createNormalizeFeature } from './normalize';
+import { ets2IsoA2 } from './populated-places';
 
 type RegionName = 'usa' | 'europe';
 
@@ -39,6 +44,8 @@ export class LabelProducer {
    */
   readonly dataProvider: LabelDataProvider;
 
+  private readonly hasMeta: boolean;
+
   /**
    * @param gameData
    *     The game map data to use as a primary source.
@@ -53,6 +60,9 @@ export class LabelProducer {
     gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>,
     metas?: LabelMeta[],
   ) {
+    metas ??= [];
+    this.dataProvider = new LabelDataProvider(gameData, metas);
+    this.hasMeta = metas.length > 0;
   }
 
   /**
@@ -64,6 +74,20 @@ export class LabelProducer {
    * @returns All map labels for the provided game data and metadata.
    */
   makeLabels(): Label[] {
+    const labelClass = {
+      usa: AtsLabel,
+      europe: Ets2Label,
+    }[this.dataProvider.region];
+    let labels: Label[] = this.dataProvider.mileageTargets.map(
+      target => new labelClass(target, this.dataProvider),
+    );
+
+    if (this.hasMeta) {
+      labels.forEach(label => this.dataProvider.assignMeta(label));
+      labels = labels.concat(this.dataProvider.missingLabels(labels));
+    }
+
+    return labels;
   }
 
   /**
@@ -81,6 +105,9 @@ export class LabelProducer {
     dir: string,
     region: RegionName,
   ): MappedDataForKeys<['cities', 'countries', 'mileageTargets']> {
+    return readMapData(dir, region, {
+      mapDataKeys: ['cities', 'countries', 'mileageTargets'],
+    });
   }
 
   /**
@@ -93,6 +120,7 @@ export class LabelProducer {
    * @see {@link LabelDataProvider.readMetas}
    */
   static readMetas(jsonPath: string): LabelMeta[] {
+    return LabelDataProvider.readMetas(jsonPath);
   }
 }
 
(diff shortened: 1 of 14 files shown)


truckermudgeon:
Some quick thoughts:

I _do_ think this new `extra-labels` module is worth having in the repo, regardless of how we end up exchanging data 🙂 

 > The primary disadvantage here is that any changes I make aren't visible on the live server until you deploy them.

If one of the requirements is that you're able to change what's shown on the live server without any action on my part, then I think we'd have to stick with what we have now: `SceneryTownSource.tsx` references a GeoJSON file you maintain, at a URL that I occasionally update every SCS content release 😬.

But if that's not a must-have requirement, then the label metadata JSON could be treated like any other `generator` resource (e.g., the ets2villages CSV and the achievements JSON). It could exist in the repo (assuming that `ats-towns` contains a label metadata JSON file, then `maps` could reference `ats-towns` as a git submodule), and it would be read during the build process. You're free to make updates to label metadata in `ats-towns`, and it's on me to update `maps`' submodule reference when this happens (just as it's currently on me to update the url in SceneryTownSource every time new DLC is released).

The online map wouldn't get the benefit of automatically incorporating any minor bugfixes you make to files at a URL you maintain... but maybe that's a minor benefit that hasn't been seen very often, anyway?

nautofon:
> If one of the requirements is that you're able to change what's shown on the live server without any action on my part,

It's not a requirement, it's a potential advantage: Combined with the new tools, such access *might* help motivate me to maintain the future dataset better than the current one.

> then I think we'd have to stick with what we have now: `SceneryTownSource.tsx` references a GeoJSON file you maintain, at a URL that I occasionally update every SCS content release 😬.

I was about to say "well actually you might not have to update the URL anymore", but I'm realizing just now that as of this moment, this is only true for `TargetLabel` instances, not for other label instances. Hmm. 🧐

>`maps` could reference `ats-towns` as a git submodule

Ah, that's another interesting option.

> The online map wouldn't get the benefit of automatically incorporating any minor bugfixes you make to files at a URL you maintain... but maybe that's a minor benefit that hasn't been seen very often, anyway?

Hah! 🤣 That's an exceedingly polite way to put it. I think that benefit has been seen a grand total of zero times so far.

But that's in part because I've been having the change offered with this PR on my mind for quite a while, and I didn't want to spend much time on the old dataset. So *maybe* things will change; we'd have to see.

nautofon:
I think you're right in that there are several important advantages to treating labels the same as any other generator resource: Fewer special cases, easier to work with in a local install, no need to update URLs, and no need to worry about [incompatible attributes](https://github.com/nautofon/ats-towns/commit/7ee62fc97c58ff5d9cb742ac17621f92d1ba8533), either.

So, how do you feel about this:  
We plan to include the metadata table here in `maps`, *once the metadata table is sufficiently complete.*

Until then, we stick to the old scheme, with the URL pointing to my repository. I don't know what *"sufficiently complete"* means yet, but we'll figure it out.

truckermudgeon:
> We plan to include the metadata table here in `maps`, _once the metadata table is sufficiently complete_.

👍 

> Until then, we stick to the old scheme, with the URL pointing to my repository. I don't know what _"sufficiently complete"_ means yet, but we'll figure it out.

👍 👍

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,891 @@
...
+
+  protected readonly gameData: MyMappedData;
+  protected readonly metaData: LabelMeta[];
+  protected readonly metaDataByToken: Map<string, LabelMeta>;
+
+  /**
+   * @param gameData
+   *     The {@link MappedData} to use as a primary source.
+   *     It's not necessary to give the full object; it's sufficient to pick
+   *     the properties `cities`, `countries`, `mileageTargets`, and `map`.
```
truckermudgeon: Soon to land on `main`: an "official" way of doing this (see https://github.com/truckermudgeon/maps/pull/48).

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,891 @@
...
+
+  /**
+   * Reads the metadata table from a file into memory.
+   *
+   * @param jsonPath - The path of the metadata file. Only JSON is implemented.
+   *
+   * @returns The metadata table as array.
+   */
+  static readMetaData(jsonPath: string): LabelMeta[] {
+    return readArrayFile<LabelMeta>(jsonPath);
```
truckermudgeon: Whoops, just noticed this call to `readArrayFile`, which won't be possible once #48 lands 😬 

Using `readArrayFile` is probably not the appropriate thing to do here, anyway, because `readArrayFile` assumes that `jsonPath` refers to a `parser`-generated file and will do things like special-case strings that look like hexadecimal bigints.

I think `LabelDataProvider::readMetaData` should instead read and parse the file contents manually, e.g.:

```ts
const json = JSON.parse(fs.readFileSync(jsonPath, 'utf-8'));
if (!Array.isArray(json)) {
  throw new Error();
}

// optionally validate that contents of json conform to the LabelMeta spec
// ...

return json as LabelMeta[];

**Comment by truckermudgeon** [comment]
I haven't started reviewing this PR in earnest (I'll wait for it to be marked as "Ready for review"), but I did leave a couple of comments related to an upcoming change to `readMapData` 🙂

File: packages/clis/generator/geo-json/extra-labels.ts (hunk 1 above)
Comment (nautofon) [reply packages/clis/generator/geo-json/extra-labels.ts]: Ah, great! 😀

File: packages/clis/generator/geo-json/extra-labels.ts (hunk 2 above)
Comment (nautofon) [reply packages/clis/generator/geo-json/extra-labels.ts]: Good point, thanks.

By the way, I might change it to be able to read CSV here. Depends on which way to exchange that data we end up choosing. (I rather like your submodule idea.)

**Comment by nautofon** [comment]
I think we're just about ready to review the proposed changes here.

There are some questions I want to ask you—things where I'm unsure about the best way of solving a problem, or maybe just about your preferences. Would you like me to ask these up front and adjust the branch accordingly before you start a review? Or would you prefer to look through everything yourself first?

**Comment by truckermudgeon** [comment]
>  Would you like me to ask these up front and adjust the branch accordingly before you start a review?

Asking up front SGTM. They could be left as inline comments or PR-level comments... and at that point, the PR's probably ready to be marked as ready-for-review :)

**Comment by truckermudgeon** [comment]
One quick (to say, but probably not to action 😅 ) question:

Is there a way to break up this PR into smaller chunks? For me, ~500 LOC is ideal (where fixtures don't count towards that LOC), but I can see how that might not be possible / easy / make sense.

Off the top of my head:
- `geo-json/extra-labels.ts` could be a separate PR, and probably can't be broken up even more
- the tests + fixtures could be a separate PR
- `commands/extra-labels.ts` could be a separate PR

The PRs could be merged one-by-one to `main`, or they could be merged one-by-one into a feature branch, and then that feature branch could be merged into `main` in one go (I have a slight preference for that latter, but I don't feel too strongly either way).

**Comment by nautofon** [comment]
For this PR, `vitest run --coverage` reports 509 LOC in `cli/generator` (excluding tests and docs).

Tests or docs in a separate PR doesn't sound right to me. I *could* split out the CLI if you need me to (167 of the 509 LOC). And the changes to `apps/demo`, but those are pretty minor anyway.

This PR is largely self-contained, with almost all changes in newly added files. I'm hoping that's easier to review than a similarly-sized PR that makes assorted changes in lots of files that already exist.

**Comment by truckermudgeon** [comment]
> For this PR, `vitest run --coverage` reports 509 LOC in `cli/generator` (excluding tests and docs).

Ah. Fair enough.

> Tests or docs in a separate PR doesn't sound right to me. 

Yeah, I did feel a little weird suggesting tests in a separate PR. Ok 👍 

> This PR is largely self-contained, with almost all changes in newly added files. I'm hoping that's easier to review than a similarly-sized PR that makes assorted changes in lots of files that already exist.

Also fair.

When it comes time to review, I will most likely end up reviewing the code in those chunks I mentioned (just setting the expectation that there _may_ be multiple rounds of multiple rounds of comments)

**Comment by nautofon** [comment]
I guess something we *could* do is split out the user-visible API and review that part first. So basically, just TSDoc comments plus some empty method bodies. (The compiler won't like it, but it doesn't have to like it.)

Would that make sense to you?

**Comment by truckermudgeon** [comment]
> split out the user-visible API and review that part first. 

That SVGTM. Settling on an API is usually the most back-and-forth-y part of a PR, in my experience.

File: packages/clis/generator/geo-json/extra-labels.ts [hunk 3]
```diff
@@ -0,0 +1,581 @@
...
+ * @packageDocumentation
+ * @see {@link LabelProducer}
+ * @author nautofon
+ */
+
+import type { City, Country, MileageTarget } from '@truckermudgeon/map/types';
+import type { GeoJSON } from 'geojson';
+import type { MappedDataForKeys } from '../mapped-data';
+
+type RegionName = 'usa' | 'europe';
```
nautofon, line 22: I strongly dislike naming this union `map`, as that term is already overloaded and there are better alternatives, like "region".

But this type alias as such I'm not sure about. After refactoring, `RegionName` is only used in two places in this module. Would it be a good idea to have a globally used type alias for `'usa' | 'europe'`, like maybe in `libs/map/types.ts`?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+
+/**
+ * A façade to this module. Offers an easy interface to create map labels
+ * for whatever combination of game data and meta data you throw at it.
```
nautofon, line 26: (Façade design pattern: Library mindset...)

It's definitely useful to have the CLI and tests share a common entry point into this module. I feel `LabelProducer` does carry its own weight, but just *barely* so, given that the `Label` and `LabelDataProvider` APIs it's hiding aren't too complex. Still, I'd *prefer* to keep this façade class.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+  /**
+   * @param __namedParameters.gameData
+   *     The game map data to use as a primary source.
+   * @param __namedParameters.metaData
+   *     The metadata table to use for augmenting the labels generated from
+   *     mileage targets in the game data. Optional.
+   *
+   * @see {@link clis/generator/mapped-data!readMapData}
+   * @see {@link readMapData}
+   * @see {@link readMetaData}
```
nautofon: I use the capitalization `metaData` in this module, for consistency with `mapData` and `gameData`. But it irks me because "metadata" is a single word, not two words. What would you do?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+   */
+  constructor({
+    gameData,
+    metaData = [],
+  }: {
+    gameData: MappedDataForKeys<['cities', 'countries', 'mileageTargets']>;
+    metaData?: LabelMeta[];
+  }) {
```
nautofon: The `LabelProducer` and `LabelDataProvider` constructors take essentially the same arguments. The significant difference is that one uses named args, the other positional args. I think both should use the same style, but which one?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+   */
+  makeLabels(): Label[] {
+  }
+
+  /**
+   * Like {@link makeLabels}, but additionally filters out all invalid labels.
+   *
+   * @returns All _valid_ map labels for the provided game data and metadata.
+   */
+  makeValidLabels(): Label[] {
```
nautofon: The CLI app ended up not actually using this method. We have tests for it, so I didn't see anything wrong with keeping it anyway. But now, after our discussion in the other thread, I recognize library mindset. 🙂

Up to you.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+
+  /**
+   * @returns The metadata for the map label as a shallow copy of this object.
+   */
+  meta(): LabelMeta;
+
+  /**
+   * @returns The map label as a GeoJSON point feature.
+   *
+   * @throws ReferenceError if no coordinates are available.
```
nautofon: Should this be a `ReferenceError` or a `RangeError`, or something else entirely?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+  /**
+   * Note or comment about the label or its attributes.
+   */
+  remark?: string | null;
+
+  /**
+   * Reference to real-life information about the labeled entity.
+   * Not currently used; reserved for future expansion.
+   */
+  ref?: unknown;
```
nautofon: `ref` is currently unused. Should it even exist as a field in implementing classes?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+   * names, but {@link clis/parser/game-files/map-files-parser!parseMapFiles}
+   * currently only provides names localized into English.
+   *
+   * @param name    - The city name to look up.
+   * @param country - The mileage target country to restrict the search to.
+   *
+   * @returns The {@link City} object. `undefined` if none was found or no
+   *     `country` was given.
+   */
+  cityFromName(name: string, country: Country | undefined): City | undefined {
```
nautofon, line 675: The 2nd argument could alternatively be defined as `country?: Country`. Is there a functional difference? Which do you prefer?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+ * It uses the template method pattern in {@link targetAnalysis} to give
+ * subclasses an opportunity for adjusting parts of that analysis to match
+ * the peculiarities of each game.
+ */
+export abstract class TargetLabel extends GenericLabel {
+  /**
+   * The original {@link MileageTarget} that this label was generated from.
+   * @internal
+   */
+  readonly target;
```
nautofon: A type annotation is optional here. TS infers the type from the assignment in the constructor. Should I include the type in the field definition anyway? (I did so in the façade class.)

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+ * All attributes are optional. When applying metadata to labels generated
+ * from mileage targets, an __undefined__ attribute (`null`) should cause
+ * that attribute to be undefined in the result as well, whereas a
+ * __missing__ attribute should cause the result to use the mileage target
+ * data for that particular attribute.
+ *
+ * @see https://github.com/nautofon/ats-towns/blob/main/label-metadata.md
+ */
+export interface LabelMeta {
+  // Meant for JSON, thus this interface must use null rather than undefined.
```
nautofon: Should the `LabelMeta` interface be defined in this file, or maybe rather in `libs/map/types.ts`? We'll need it in the demo app.

`libs/map/types.ts` is already somewhat convoluted, and importing the interface from here isn't *too* bad, so I'd tend towards keeping it here:

```ts
import type { LabelMeta } from '@truckermudgeon/generator/geo-json/extra-labels';
```

If we *do* move it, we'd need to consider whether `Label` below should be moved as well.

nautofon:
Done. The downside of this approach is that GH carps a bunch of CI errors into the source view. ~I guess we'll have to live with that.~   _actually, [you can hide them](https://github.com/user-attachments/assets/45635038-ae57-48a2-9f80-695849d2c774) #TIL_

This is now "ready for review", but I won't click that button until this branch is complete again. 🙂

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: > Would it be a good idea to have a globally used type alias for `'usa' | 'europe'`, like maybe in `libs/map/types.ts`?

It would 😁. I can update existing code to make use of it after all the `extra-labels` work lands.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: I'd make it read `readMetadata`, with the reasoning that the "names should be camel cased based on words" convention overrides the "other uses of xxxData have Data capitalized" ~convention~ coincidence.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: I'd prefer positional, because of some vague rule in my head that I can't quite explain yet (and which I'm sure I've probably violated in the codebase).

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: I'd leave it out for now. I'd rather keep the API surface small, and only add methods like this if callers show a strong need for it (e.g., they would only ever call this over `makeLabels`).

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+   * Like {@link makeLabels}, but additionally filters out all invalid labels.
+   *
+   * @returns All _valid_ map labels for the provided game data and metadata.
+   */
+  makeValidLabels(): Label[] {
+  }
+
+  /**
+   * Reads game data from the parser output into memory.
+   * Suitable for feeding into the {@link LabelProducer} constructor.
```
truckermudgeon, line 96: 🤔 I'm unsure of the benefit of this static `readXXX` methods as part of the `LabelProducer` class.

If it's likely that this will only be called to construct a `LabelProducer`, then could this be either replaced or supplemented with, the following?

```ts
static fromData(pathToParserOutput, pathToMetaJson, region): LabelProducer {
  return new LabelProducer(...);
}
```

(And if consumers need more control over the data sources of a LabelProducer, e.g., for tests, they can call the ctor directly and provide their own `MappedData` and `LabelMeta[]` instances.)

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: 🤔 initial thought is to move it to `libs/map/types.ts`, then deal with that mess later / separately. Seeing imports to the generator CLI from anything outside of the generator CLI feels a little weird. 

I'll continue to give this some thought. The one thing I want to avoid is: the packages in `maps` get published to `npm`. A user who wants to download the Demo-app package has to download the generator-cli package, too... even though they will never ever call the generator cli.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: There is a functional difference: see [this playground link](https://www.typescriptlang.org/play/?ssl=10&ssc=1&pln=10&pc=22#code/GYVwdgxgLglg9mABMOcAUBDATgcwAwBciAzlFjGDgDSLY4CMRYIAtgEYCmWiAPouABMOwChwEBKRAG8AUAF8ZM0JFgJEbbJlyESZCtVq56AfiatOWSbIWKU6AOQALGPckB6N4i5Y43AO6OHEh09IgwxGD2UCQADhwQMCJiihpYaE4u7p5gcF5YPlhAA).

If it's likely that callers will not have a `Country` value to provide when calling this, or specifying a `Country` only serves to enhance the lookup behavior, then I'd prefer `?` because it seems like it's truly optional. 

But if callers are expected to provide both a city name and `Country`, and it's just that specifying an `undefined` value for `Country` will always result in an `undefined` `City` being returned, then I'd say there's meaning/value to make callers aware of this, and so I'd opt for the current `| undefined`.

> ```
> * @returns The {@link City} object. `undefined` if none was found or no
> *     `country` was given.
> ```

And just to confirm: if no `country` is given, then this always returns `undefined`? If that's the case, and callers are expected to provide both a city name and `Country` (even one that's `undefined`), then this can be overloaded so that there's a stronger type guarantee, something like:

```ts
cityFromName(name: string, country: Country): City | undefined;
cityFromName(name: string, country: undefined): undefined;
cityFromName(name: string, country: Country | undefined): City | undefined {
  // ...
}
```

and then TS will know that if you call `cityFromName(name, undefined)`, the return type is always `undefined`.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+  /**
+   * Search for a city in a particular country's game data by city token.
+   *
+   * @param token   - The city token to look up.
+   * @param country - The mileage target country to restrict the search to.
+   *
+   * @returns The {@link City} object. `undefined` if none was found or no
+   *     `country` was given.
+   */
+  cityFromToken(token: string, country: Country | undefined): City | undefined {
```
truckermudgeon, line 695: Same comment as above

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+  /**
+   * Search for a country in the game data by ISO 3166-1 or ISO 3166-2 code.
+   *
+   * @param isoCode - The country or country and subdivision code to look up.
+   *
+   * @returns True if the given code identifies a country in the game data.
+   *
+   * @see {@link clis/generator/geo-json/populated-places!ets2IsoA2}
+   */
+  isValidCountry(isoCode: string): boolean {
```
truckermudgeon: Given its input and the meaning of its output, WDYT of:

```ts
isKnownCountryCode(isoCode: string): boolean
```

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+  isInRegion(meta: LabelMeta): boolean | undefined {
+  }
+
+  /**
+   * The name of the region for which game data is provided.
+   *
+   * @returns `'europe' | 'usa'`
+   */
+  region(): RegionName {
+  }
```
truckermudgeon: If the return value doesn't change during the lifetime of a `LabelDataProvider`, can this be a `readonly region: RegionName` field?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+ * needed for mileage target analysis.
+ */
+export class LabelDataProvider {
+  // This class is somewhat inefficient, with nested sequential searches.
+  // But we have at most a few thousand entries, so it doesn't really matter.
+
+  /**
+   * The full list of {@link MileageTarget} objects read from the parser output.
+   */
+  readonly mileageTargets;
```
truckermudgeon: For field declarations, could you specify the type, e.g., `readonly mileageTargets: readonly MileageTarget[]`?

Avoids having to read the ctor implementation to figure out what kind of data this stores.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+
+/**
+ * Details on the mileage target analysis results.
+ * @internal
+ */
+export interface TargetAnalysis {
+  /**
+   * Mileage target name, tidied for use as map label text.
+   */
+  tidyName: string;
```
truckermudgeon, line 483: Can these fields be marked as `readonly`? Maybe it doesn't matter too much, since this is marked as `@internal` 🤔

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: Oh, just noticed this _after_ making a comment about this 😅 

Yes, I'd prefer type annotations for field declarations.

The only time where I wouldn't say anything is in something like:

```ts
readonly someMap = new Map<string, number>();
```

because it's easy to see the inferred type of the field there.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+ */
+export interface LabelMeta {
+  // Meant for JSON, thus this interface must use null rather than undefined.
+
+  /**
+   * The token identifying the mileage target to apply the label attributes to.
+   *
+   * If missing or undefined, this object describes a new label instead.
+   */
+  token?: string | null;
```
truckermudgeon: This may be the meatiest comment:

Should these be declared as `readonly`? Can a `LabelMeta` be treated as something immutable (because as a consumer, there's less to worry about / question when you're working with immutable objects).

Or is mutability something essential to `LabelMeta`-ness?

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+   * @returns True if the map label is considered valid for display.
+   *
+   * Examples for map labels that aren't valid:
+   * - No coordinates are available.
+   * - No label text is available (e.g. feature kind is `unnamed`).
+   * - No country / state could be determined (may indicate the label is
+   *      for a location in a DLC that hasn't been released, or that it's
+   *      located in a different game).
+   */
+  isValid(): boolean;
```
truckermudgeon: Hmmm... I guess this being a method instead of a field is because `LabelMeta` is mutable, and whether or not it's valid can change depending on what happens to those mutable values.

If `LabelMeta` becomes immutable, then this could be declared as `readonly isValid: boolean`.

packages/clis/generator/geo-json/extra-labels.ts
```diff
@@ -0,0 +1,581 @@
...
+
+  /**
+   * @returns The metadata for the map label as a shallow copy of this object.
+   */
+  meta(): LabelMeta;
```
truckermudgeon: Same comment as above.

But also: it's a little surprising to see something like:

```ts
class Foo extends Bar {
  bar(): Bar
}
```

because if I have a Foo, and I want a Bar... well, I already have a Bar.

I think this is related to the conversation we had about destructuring: we want an object with certain fields, only. So having a method that creates such an object makes sense.

And we mention that the implementation produces a shallow copy, because callers may want to modify fields, and we don't want those modifications to affect the source Label instance.

WDYT of renaming this to something that has the word "copy" or "create" in its name, so the shallow-copy-ness is more upfront? First things that comes to mind: `toMetaCopy()` and  `createLabelMeta()` (naming is hard 🙃).

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: I haven't made use of built-in Error types before... but after reading up on ReferenceError and RangeError on MDN, they both sound inappropriate. I think plain ol' Error is good enough here...

...but I'm wondering: when would no coordinates be available? Is that only the case if this Label is invalid? If so, then sounds like the implementation should call `Preconditions.checkState(this.isValid())` as a signal that the caller should only call this method when the Label is valid.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: I'm thinking, "no".

It can always be added if/when it needs to be.

truckermudgeon (commented):
> actually, [you can hide them](https://github.com/user-attachments/assets/45635038-ae57-48a2-9f80-695849d2c774) #TIL

Ha, I had a comment and screenshot all prepped for this, but you beat me to it 🙂 

Comments in. Apologies if they're a little all over the place: I reviewed before work, during lunch, and after work  😅

packages/clis/generator/geo-json/extra-labels.ts
nautofon: All right, then I'll leave this type alias in. What about the name?

As you know, the equivalent CLI option is currently named `--map`. It'd be nice to use similar terms for the type and the CLI option. If that option were to be renamed, `--map` could become like a deprecated alias or something.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: I'd be okay with that, but I'm just realizing that I simply call it `meta` pretty much everywhere else. So that method should perhaps just be called `readMeta()` instead—or maybe the slightly awkward plural form `readMetas()`, as it returns an array.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Works for me. 🙂

(FWIW, I generally prefer named args in situations like this, but didn't find a convincing way to write them in TS.)

packages/clis/generator/geo-json/extra-labels.ts
nautofon: I'm not entirely sure what you're asking here.

* These methods in `LabelProducer` are just aliases for their equivalents in `LabelDataProvider`. I don't mind getting rid of these aliases. Doing so would mean one additional import for clients, which might go against the spirit of the Façade pattern, but I don't think it'd be a big deal.
* The `readMeta()` method may look simple at the moment. However, in future I expect we'll expand it to read CSV and directories in addition to JSON. That'll be more than enough complexity to justify the method, and providing it already now hides those details from clients.
* Thanks to your work on `MappedDataForKeys`, the `readMapData()` method has indeed become very simple. It's useful not to force clients to think about which keys to pick. There are probably other ways to do that than this method. But I do think the method makes sense, if only for symmetry with `readMeta()`.
* Replacing public constructors with static factory methods is a common pattern. But I'm not sure what its advantage would be in this case.
* The specific `fromData()` method you're suggesting here wouldn't be usable by the current implementation of the CLI, which needs to read game data and metadata separately. So that method would be unused, similarly to `makeValidLabels()`.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Agreed on `LabelMeta`: It does feel a little weird. But I think it makes sense to keep `Label` here. `Label` is more closely related to the classes in this module, and it currently isn't needed by the demo app.

I don't understand your comment about npm. Maybe it wasn't meant for me? 🙂

packages/clis/generator/geo-json/extra-labels.ts
nautofon: > Or is mutability something essential to `LabelMeta`-ness?

Excellent question! Truthfully, it's not something I even considered here. I took a day to think it over, but without really finding a great answer.

The build process is a multistep operation, including things like `assignMeta()` in the data provider. So while creating a label, it should be modifiable. Could the design be changed to e.g. cast a label into an immutable type once creation has completed? Yeah, sure. But I'm not convinced it would make sense to do that.

The question fundamentally is: Should a change—any change—to a map label cause the *identity* of that label to change? (An immutable implementation prohibits modifying a label, thus requiring the label to be replaced with a new one.)

The clients we have right now (CLI and demo app) have ownership of the labels and no need to change them at all, so there really are no reasons to worry in the first place. Accordingly, from an "app mindset" point of view, the question of mutability doesn't matter, and we should just do the simplest thing, to ease future maintenance.

But with library mindset, there might well be other clients in future that don't exist yet, like maybe a web app for editing labels. I don't have plans to work on anything like that, but who knows what happens. My intuition is that immutable labels would make writing such an app harder, not easier. But not being well familiar with the 
TS ecosystem, I might just be wrong about that.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: `Label` and its implementing classes define the business logic of map labels. The `isValid()` method is where the determination is made whether a label is valid for being usable on the map.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: The method does return with the same (super-)type, but with a different value. That's not exactly uncommon. By itself, I wouldn't view this as a smell.

But by making your question about `extends` as well, you've exposed a design error. 👍 That keyword defines an "is a" relationship between types. So what we're actually saying here is that "a label is metadata". That doesn't sound quite right.

Instead of "is a", we should be using a "has" relationship—in other words, composition rather than inheritance. With that, `meta()` shouldn't look surprising anymore.

Thanks for catching this!

```patch
-export interface Label extends LabelMeta {
+export interface Label {
```

packages/clis/generator/geo-json/extra-labels.ts
nautofon: `Error` is a built-in error type, too… 🙂 I do agree with you that the other two don't fit very well here. I guess what I meant to ask was: What's the best way to signal this particular error?

I can't seem to find the docs for `Preconditions`. Could you give me a pointer?

packages/clis/generator/geo-json/extra-labels.ts
nautofon: `TargetAnalysis` is designed as a mutable type.

It's a purely internal data structure which is only exposed to clients for debugging and analysis purposes, and only *after* their internal use is completed. So at that point, member visibility doesn't matter anymore.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Thank you for the clear explanation, and for reminding me of signature overloads! The overloads you show above are exactly what we need here.

`country` is indeed not optional, that's why `?` didn't feel quite right to me, either—even though it'd be shorter.

The reason `country` is required for `cityFromName()` is to reliably distinguish between e.g. Sidney, MT and Sidney, NE. If the country or state isn't actually known for whatever reason, I think it's safest to not match a city at all, rather than produce output that might be wrong. The same goes for `cityFromToken()`, but only as a sanity check to defend against city tokens that exist in both games, of which I think there aren't any.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Ahyes, that's better.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: It can be, sure. But I doubt it makes sense. We can discuss that in https://github.com/nautofon/maps-nodejs/pull/1 later.

We're currently discussing the interface, so I *think* what you're really asking is: Should clients call this with `dataProvider.region` or `dataProvider.region()`? That's actually closely related to something I wanted to ask you about, but forgot. 🙂

Setters are lvalue methods. Once you use setters for something, you also need getters. Fine.

But what if something isn't meant to be changed? Limiting getters to be *only* used when you also have setters doesn't sound right to me. So what are the situations where you would use getters instead of regular methods in TS? Is it just a matter of preference?

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Oh, sorry, I skipped over your question before. Coordinates are unavailable:
* for labels created from mileage targets that have neither `node_uid` nor `position`,
* for labels created from mileage targets with a `node_uid` that the parser couldn't resolve to a node,
* for labels created from mileage targets with metadata that says the target's coordinates are wrong and to be deleted, and
* for labels created from metadata records (instead of mileage targets) that lack coordinates.

It's a rare condition. Maybe 1% of all labels are affected.

nautofon:
I left some replies. GH makes a bit of a mess of them here in the "conversation" tab. 😕 Might be easier to read in the "files changed" view...

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: `RegionName` SGTM

>  the equivalent CLI option is currently named `--map`. It'd be nice to use similar terms for the type and the CLI option.

Agreed. Can rename (or deprecate and add new option) once the new type alias lands.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: `readMeta()` or `readMetas()` SGTM (I have a slight, but not strong, preference for the latter).

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: > I can't seem to find the docs for `Preconditions`. Could you give me a pointer?

Whoops, I never added comments to [precon.ts](https://github.com/truckermudgeon/maps/blob/main/packages/libs/base/precon.ts) 😬 

That class is loosely based on Guava's [Preconditions](https://guava.dev/releases/snapshot/api/docs/com/google/common/base/Preconditions.html).

Relevant quotes:

> Static convenience methods that help a method or constructor check whether it was invoked correctly (that is, whether its preconditions were met).
>
> If the precondition is not met, the Preconditions method throws an unchecked exception of a specified type, which helps the method in which the exception was thrown communicate that its caller has made a mistake. 
>
> It is of course possible to use the methods of this class to check for invalid conditions which are not the caller's fault. Doing so is not recommended because it is misleading to future readers of the code and of stack traces.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: > If the country or state isn't actually known for whatever reason, I think it's safest to not match a city at all, rather than produce output that might be wrong.

👍 

> The same goes for `cityFromToken()`, but only as a sanity check to defend against city tokens that exist in both games

👍 . I'm all for checking sanity 🙃

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: Agreed on keeping `Label` here.

About npm: that comment is more for me, and documenting my thoughts 😅. Feel free to ignore.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: > These methods in `LabelProducer` are just aliases for their equivalents in `LabelDataProvider`

Is `LabelDataProvider` meant to be used by anything else other than `LabelProducer`? Maybe it is, because it's `export`ed?

If it isn't, and `LabelDataProvider` is just a `LabelProducer` implementation detail, then exposing aliases into that implementation detail seems a little odd. I think it's fine for `LabelDataProvider` to have `readMeta` and `readMapData` methods, but do consumers of the `LabelProducer` facade need to know about them?

> The specific `fromData()` method you're suggesting here wouldn't be usable by the current implementation of the CLI, which needs to read game data and metadata separately.

Ahhh... that's what I was missing. Looking ahead:

```ts
  const metaRecords = metaPaths.flatMap(p => LabelProducer.readMetaData(p));
  const labels = args.map.flatMap(region => {
    const mappedData = LabelProducer.readMapData(args.inputDir, region);
    const producer = new LabelProducer({
      gameData: mappedData,
      metaData: metaRecords,
    });
    // ...
```

Metadata needs to be read first, because
- one metadata file can apply to multiple regions
- multiple metadata files can apply to a single region

🤔 It'd probably be too awkward / not worth it to require a 1:1 pairing of metadata files to regions in the CLI. So keeping the static `readMeta` and `readMapData`  SGTM.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: >  I _think_ what you're really asking is: Should clients call this with `dataProvider.region` or `dataProvider.region()`

Yes.

> Setters are lvalue methods. Once you use setters for something, you also need getters.

It's possible to define an Object with a setter for a property, without defining its getter (which is weird and probably isn't ever done in real code... but that's probably not that important for this discussion 😅) 

> But what if something isn't meant to be changed? 

If an interface defines something that isn't meant to be changed by callers, then I'd _expect_ that something to be defined as a `readonly something: SomeType`. I wouldn't be surprised / confused if the interface defined it as `something(): SomeType`, but I'd think, "that doesn't seem JS-y / TS-y to me".

> Limiting getters to be only used when you also have setters doesn't sound right to me.

Doesn't sound right to me, either. Getters should be usable anytime you want to expose a way to get a value.

> So what are the situations where you would use getters instead of regular methods in TS? Is it just a matter of preference?

For me, I've built up a preference for getters over regular methods because it seems to be the convention in JS/TS. Some examples:
- querying the length of a JS array is done with `someArray.length`; the size of a map is done with `someMap.size`
  - `Array::length` is a getter and setter. `Map::size` is just a getter.
- a DOM Node has a bunch of properties, e.g., [`Node::firstChild`](https://developer.mozilla.org/en-US/docs/Web/API/Node/firstChild) and [`Node::textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent). Some of which are read-only, some of which aren't.
- Not sure what's going with the NodeJS APIs, but some objects have read-only properties (like [`dirent::name`](https://nodejs.org/api/fs.html#direntname)), while others have methods (like [`dirEnt::isdirectory()`](https://nodejs.org/api/fs.html#direntisdirectory). I'm not sure what the reasoning is.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: > The clients we have right now (CLI and demo app) have ownership of the labels and no need to change them at all, so there really are no reasons to worry in the first place. Accordingly, from an "app mindset" point of view, the question of mutability doesn't matter, and we should just do the simplest thing, to ease future maintenance.

Leaving things as mutable SGTM. I mostly brought it up because:

> future maintenance

I've been bitten in the past by mutating things returned from APIs... but that was because I was writing spaghetti code.

I guess as a consumer, if I had a personal preference for working with immutable types, then I could always type any `LabelMeta`s returned from these APIs as readonly... no need to mandate that at the API level.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: Mutability aside, this is probably more related to the question of property getters vs. methods.

Validity seems to be a property of a `LabelMeta`. So why not define it as a property instead of a method?

Or is there something about validity that is better communicated with a method? Maybe it's a non-trivial calculation, so shouldn't be seen as a cheap property access?

truckermudgeon:
Comments in. I am now agonizing over properties vs. methods, and when to use one over another.

I guess in the end, it doesn't really matter, and I don't want to hold up the PR train unnecessarily.

My weakly-held current thinking is:
- if the value of the thing-returned doesn't change over the lifetime of an object, then use a property
- if the value of the thing can be get and set, use properties
- if the value of the thing-returned can't be set and the thing-returned is produced through some non-trivial calculation, use a method
- otherwise, use a property

packages/clis/generator/geo-json/extra-labels.ts
nautofon: > Is `LabelDataProvider` meant to be used by anything else other than `LabelProducer`?

Sure, clients can use it directly if they need to.[^1] It's not an implementation detail, it's just a lower level.

The aim of the Façade design pattern is to offer a higher-level API for clients who don't need the flexibility a lower-level API offers. It does so by delegating the work down to the lower-level classes, and sometimes that's done by providing an alias.

> I think it's fine for `LabelDataProvider` to have `readMeta` and `readMapData` methods, but do consumers of the `LabelProducer` facade need to know about them?

Not as long as the aliases in `LabelProducer` exist, no. If we removed these aliases, clients would need to additionally import `LabelDataProvider` to be able to use those methods instead.


> Metadata needs to be read first, because
> 
> * one metadata file can apply to multiple regions
> * multiple metadata files can apply to a single region

Right. (Perhaps the documentation should have spelled it out as clearly as that.)

[^1]:As you know, this module is primarily aimed at the ATS part of the map. A subclass for ETS2 mileage targets is provided, too. But these classes won't necessarily work well for mods like Reforma or ProMods. If somebody wanted to expand your map *privately* to cover such mods, they might need new `TargetLabel` subclasses, and at that point direct access to `LabelDataProvider` is needed, too.

nautofon: Come to think of it, I'd say we should remove `readMapData()` from `LabelDataProvider` entirely. The custom code I had in there is gone, it's just an alias to your function in `mapped-data` now. Let's keep `readMapData()` in `LabelProducer` *only.*

`readMeta()`, on the other hand, is specialized for labels. It does belong in `LabelDataProvider`, and I'd keep the alias in `LabelProducer` as well.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Ahh, it's defined locally! That explains why I couldn't find a `Preconditions` class offering a `checkState()` method on npm. My mistake. 😄

The error type I'd really been looking for earlier was Java's `IllegalStateException`, or rather a TS equivalent for it. So yes, `checkState()` looks entirely appropriate here.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Thanks for these pointers! The example of the DOM in particular demonstrates that property access is a normal way of doing business in JS.

Let's change this to `dataProvider.region` then and see how that looks.

FWIW, the reason I ended up with `dataProvider.region()` initially for the PR was simply because this would have been my *only* use of a getter. I wasn't bold enough for that and just stuck to regular methods for consistency…

packages/clis/generator/geo-json/extra-labels.ts
nautofon: > Validity seems to be a property of a `LabelMeta`. So why not define it as a property instead of a method?

*\*rationale for field vs. method; see following comments for clarification*

Clients at all times need a way to know whether a label is or isn't valid. The logic for deciding that needs to live somewhere. The only two possible options are:
* a method called by the client to check validity on demand, which could optionally cache the result of an expensive calculation internally, or
* the object constructor, which would set a ~property~ <ins>field</ins> that clients could read.

Between these options, the object constructor one requires subclasses to remember to correctly set the property. This is a potential source of bugs. It also increases the constructor's complexity, which is traditionally an anti-pattern.

By contrast, the method option keeps the code related to validity in a single obvious place. It frees subclasses from having to consider the concept of validity, while still allowing them to override the method in case they do need to change the validity value for some reason. Therefore, the method option is to be preferred, even in an immutable design.

Whether you use a real method or a getter method is another question. This situation *feels* more like a regular method to me than a use case for a getter, but perhaps I'd feel differently with more experience in modern JS. I don't have strong opinions about method vs. getter, but this should *not* be a ~regular property~ <ins>field</ins>.

nautofon:
Regarding properties vs. methods vs. getters: As far as this series of PRs is concerned, it might be more productive to consider that question once we're looking at the implementation together. Let me put these changes in and update the branches, then we can pick it up again in the next PR.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: It sounds like we're aligned on a lot of things (e.g., I see complex ctors as an anti-pattern; subclasses being allowed to override things is good), but just to clear up any confusion:

> I don't have strong opinions about method vs. getter, but this should _not_ be a regular property.

What do you mean by, "regular property"?

In my mind, getters and properties are the same thing: at the interface level, there's no difference. 

An interface with a `readonly someField: SomeType` can be satisfied by a class with a `get someField(): SomeType` getter, _or_ by a class with a `readonly someField: SomeType` field. See [this playground link](https://www.typescriptlang.org/play/?ssl=23&ssc=1&pln=23&pc=2#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHJQRwAm6IANgJ7IDO6AthAHICujARtAFzIgduUANy4AvrlwJKcOnTSZUwCJVIBJRgAdKyYFsoRm4eRix5CREuSq0Gzdl179B0UfkIIKdMFDYIw6FAAFACUOO4WBGAAFsB0AHR2rC5QyAC8yACsohYSElIycgroAApQ6JrQYNQa2rr6hhDGxeEEAOYQYPRMyY7BIXwCfa2ExGBsUCBZOch5ktKy8gBCcKkQAB6QIKQmmGUVVTX6IwToAG7QUMCkKB1dSQ5CoYMpJ5FjE1MAbDMEeUA).

nautofon: > What do you mean by, "regular property"?
> 
> In my mind, getters and properties are the same thing: [...]
> An interface with a `readonly someField` [...]

Ugh. My apologies: I confused the term "property" with "field". 🙄 Thank you for setting me straight.

(Plus, I hadn't realized that a property defined in an interface could be implemented with a getter, although that makes perfect sense.)

- - -

> Validity seems to be a property of a `LabelMeta`. So why not define it as a property instead of a method?

Other than that vague "feeling" I mentioned in the post above, I couldn't tell you. But you seem to have the opposite "feeling", and I trust yours more than mine. 🙂 Will change this to a property (implemented as getter), then we can look at it again in the implementation PR.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: The problem though is that from the callers' perspective, the type of `country` isn't known at compile time, so TS can't know which overload to resolve the call to.

Leaving this alone for now.

packages/clis/generator/geo-json/extra-labels.ts
nautofon: Changing this to `hasKnownCountryCode(meta: LabelMeta): boolean`, which I think is better still.

packages/clis/generator/geo-json/extra-labels.ts
truckermudgeon: Ahhhh... right. The only way to avoid the compiler error is to [narrow the type of `country`](https://www.typescriptlang.org/play/?#code/FAFwngDgpgBAwgS3DAvDA5AYyWdBuUSWOAewFcA7EAJzFQ03Ktv2GADNLMQESKZs4AGLUSAWwByAQzFQAFBRlQAXDADONBBQDmAGgFMaYVaUpGAlCZwwAPjEoATKOy1QHBThW69+gsCPFpWQUlVQ1qLT0DM1pVR2dXB0t7CicXCjcPLh4+ARwAySUQ2TDNHX1GGON4Q1pbFLTE5MRkO3j0txgAb2AYGBAAC1EAdxgM0YBRalFqOXMCAF82RgoNGCl6PwKg+XQBhHR9dqaCFbWAI0380ULgvYP9LFrceeW+NbEpMHOoU2Zqv5GerHDIOegAWSkgwAdNQpKlxHMYAA+GAABmhAFYYAB+BjPdAwOKpBKg07vEACK7CG47OT3Q4wT7fX7PeYwAD0HJg7CkCAANmp9D9MFIyGpYAADZk-QG0SXRflgn4NUluaFsBDsGByGWsqowACEaAoZH5-PM3V6BlWlIcAEZqf5aUUGfo9XKwOyuTBJQ6FQg1PBrG0SR0HMAFjAoILYD0+mc7QAmJ3bV37Rketl4Tncv1JgNBkFuSNsH0UEgGMQQAVQajR6YkajCsiU-kkEgAayDagF-LAGsT0foWYN1rxWxddwz7q+srZMGtqkngXTDyZc-1-3mQA) with a conditional... which seems like overkill (especially if you're likely to check if the returned `city` is `undefined`, anyway).

truckermudgeon (approved):

nautofon:
It occurred to me that you might not have gotten a notification about https://github.com/nautofon/maps-nodejs/pull/1 being "ready for review"...?

I'm in no hurry at all, simply thought I'd give you a heads-up here just in case.

nautofon:
https://github.com/nautofon/maps-nodejs/pull/2 and https://github.com/nautofon/maps-nodejs/pull/3 are now both ready for your review. 🙂

nautofon:
Can we exempt the submodule path `packages/clis/generator/resources/extra-labels` from the linter?

I see the other submodule `libdeflate` is already in `.prettierignore`. Adding `extra-labels` to it.

truckermudgeon (approved):
LGTM!

Before merging, a reminder that you'll either need to first rebase onto `main`, or merge `main` into this branch.

nautofon:
No problem. I usually rebase anyway, but I assumed you would squash this PR into a single commit on `main`, for which a rebase wouldn't have been necessary.

Would you like me to clean up the commit history of this branch?

truckermudgeon:
>  I assumed you would squash this PR into a single commit on `main`, for which a rebase wouldn't have been necessary.

🤦 you're right. There was something I saw in the UI while reviewing your PR that made me think it _was_ necessary, but I don't remember what it was exactly (and I could've misread it).

Either way, thanks for rebasing. Squashing now 🙂

PR #48, opened 2025-02-12
generator: update `readMapData` API with required `mapDataKeys` option

truckermudgeon:
This PR updates `readMapData` so that callers must specify which map data keys it's interested in via the `mapDataKeys` option, e.g.:

```ts
const mapData = readMapData(dirWithJsonFiles, 'usa', {
  mapDataKeys: ['nodes', 'roads', 'prefabs'],
});
console.log(Object.keys(mapData)); // prints out: [ 'foo', 'bar' ]
```

This means that:
* `generator` commands that only need to read a subset of the `parser`-produced JSON files will only ever read that subset[^1] of files.
* `readArrayFile` no longer needs to be called by any code outside of `readMapData`.

[^1]: well, it'll read the specified subset of files, _and_ the `cities.json` file, because I've yet to refactor the code so that reading `cities.json` is only done when absolutely necessary.


packages/clis/generator/commands/graph.ts
```diff
@@ -63,7 +63,11 @@ export const builder = (yargs: Argv) =>
 export async function handler(args: BuilderArguments<typeof builder>) {
   // TODO read only the files necessary
```
truckermudgeon:
```suggestion
```

PR #49, opened 2025-02-15
Use full width of wide terminals

nautofon:
By default, `yargs` artificially limits the usage output to 80 columns.

That wouldn't be too bad if it properly reflowed content, but AFAICT proper wrapping isn't currently possible with `yargs`, which results in line breaks in the middle of words and misaligned whitespace, for example like this:

```
  -y, --analysis   Instead of GeoJSON, write out extra-labels.json with metadata
                    for all targets. Default metadata will not be used. Useful f
                   or manual checks when updating metadata.
```

 The least we can do is use the entire width of the terminal window, minimizing the number of line breaks.

### DIFF
--- a/packages/clis/generator/index.ts
+++ b/packages/clis/generator/index.ts
@@ -15,6 +15,7 @@ import * as spritesheet from './commands/spritesheet';
 
 async function main() {
   await yargs(hideBin(process.argv))
+    .wrap(yargs().terminalWidth()) // Use full width of wide terminals.
     .command(map)
     .command(prefabCurves)
     .command(cities)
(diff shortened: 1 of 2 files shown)


truckermudgeon (approved):
> AFAICT proper wrapping isn't currently possible with `yargs`

I dug a little into this and found https://github.com/yargs/yargs/issues/2112. It looks like `yargs` _can_ do smarter wrapping at word boundaries, but that's only available when importing `yargs` as a CJS module (`maps` uses ES modules). 

Ugh. I don't know the details of why there are multiple module system standards... I just know I'm annoyed by it 🙃 (I've run into issues with other dependencies because of the whole CJS vs. ESM thing).

Back to this PR: I run a 180-column wide terminal, and this PR is an improvement. Thanks for raising it 🙂

PR #50, opened 2025-02-15
Consider tuning the linter configuration to better match this project's needs

nautofon:
Being new to TypeScript, I feel a bit weird even *suggesting* to change your linter config.

But I see in [eslint.config.mjs](https://github.com/truckermudgeon/maps/blob/0e55b84b35680c0b6cc8520752986fcee4631eec/eslint.config.mjs) that `maps` is configured to largely use the "recommended" rulesets, and I know from experience with other languages that such recommendations tend to be opinionated, often being more reflective of the personal preferences of the critic / linting tool's maintainers than of general consensus on good coding style. And in the end, it's impossible for me to know how much thought you have put into the eslint config for `maps` already, so here goes nothing.

Please treat this as a request for discussion at least as much as a request for a change. 🙂

### 1. Removing properties from objects using destructuring assignments

I propose to set the [`ignoreRestSiblings: true`](https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings) option on `@typescript-eslint/no-unused-vars`.

```ts
meta(): LabelMeta {
  const { data, ...meta } = this;
  return meta;
}
// warning  'data' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

The above is from [geojson/extra-labels.ts](https://github.com/truckermudgeon/maps/blob/216c0a85aeb333d9e41196255412a0d914abf977/packages/clis/generator/geo-json/extra-labels.ts#L311-L315), where I've been struggling to find a good solution to avoid this warning. I can see several alternative ways to do this and avoid that warning, but I find the above the easiest to read by far.

It seems to me that a destructuring assignment like that is actually a good idiom for removing properties from an object, and that the linter shouldn't issue warnings for it.

### 2. Intentionally not using function arguments

I propose to set the [`argsIgnorePattern: '^_'`](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) option on `@typescript-eslint/no-unused-vars`.

```tsx
{lane.branches.map((branch, _branchIndex) => {
  return (
    <Stack>
      <div>To Node {branch.targetNodeIndex}</div>
    </Stack>
  );
})}
// warning  '_branchIndex' is defined but never used  @typescript-eslint/no-unused-vars
```

The above is from [LaneControl.tsx](https://github.com/truckermudgeon/maps/blob/0e55b84b35680c0b6cc8520752986fcee4631eec/packages/apps/prefabs/src/LaneControl.tsx#L25-L31), which is one of several spots in `maps` where existing code is already written as if the `argsIgnorePattern` option was set. This particular example is the one that has been [showing up in CI testing](https://github.com/truckermudgeon/maps/pull/49/files#annotation_32004122824).

To balance the risk of not using arguments *accidentally*, I propose to additionally set the [`reportUsedIgnorePattern: true`](https://eslint.org/docs/latest/rules/no-unused-vars#reportusedignorepattern) option. In the example above, this option would yield warnings if you did in fact use `_branchIndex` in that function.

This combination of `argsIgnorePattern` and `reportUsedIgnorePattern` gives us the flexibility to not use certain function arguments, but forces us to be explicit about our intentions.

### 3. Allowing getters as well as fields to define literals as class properties

I propose to disable [`@typescript-eslint/class-literal-property-style`](https://typescript-eslint.io/rules/class-literal-property-style/).

This rule is used to prohibit one of the two ways to declare constant literal values on classes (getters or fields), in an effort to enforce consistency throughout the entire project. But, unlike what the description of the rule implies, the two ways are not just a style variation: They also differ in how they can be overridden by subclasses.

For example, with a field foo in `class A { foo = 1 }`, `class B extends A { foo = 2 }` won't change the value of foo early enough to be visible in the superclass constructor. If  foo was a getter, that'd be different.

The best course of action consequently seems to be to generally allow both ways. That would give us the option to use the strengths of fields as well as getters as appropriate in each situation where class literal properties are used (they're hardly used in `maps` so far anyway).

- - -

There may be more potentially useful changes. But these three are the ones I noticed so far.

What are your thoughts?

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -54,7 +54,15 @@ export default [
 
     rules: {
       '@typescript-eslint/consistent-type-imports': 'error',
-      '@typescript-eslint/no-unused-vars': 'warn',
+      '@typescript-eslint/no-unused-vars': [
+        'warn',
+        {
+          args: 'all',
+          argsIgnorePattern: '^_',
+          ignoreRestSiblings: true,
+          reportUsedIgnorePattern: true,
+        },
+      ],
     },
   },
 ];
--- a/packages/apps/prefabs/src/Details.tsx
+++ b/packages/apps/prefabs/src/Details.tsx
@@ -20,7 +20,6 @@ export const Details = ({ prefab, locations }: DetailsProps) => {
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { token, path, ...prefabRest } = prefab;
   return (
     <Tabs defaultValue={0}>
--- a/packages/clis/generator/geo-json/map.ts
+++ b/packages/clis/generator/geo-json/map.ts
@@ -1182,7 +1182,6 @@ function roadToFeature(
   road: Road,
   roadLook: RoadLook,
   nodes: ReadonlyMap<bigint, Node>,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _dividerFeatures: GeoJSON.Feature<GeoJSON.LineString>[],
 ): RoadFeature[] {
   const startNode = Preconditions.checkExists(nodes.get(road.startNodeUid));


nautofon:
Converting to draft because of new warnings `Unused eslint-disable directive`.

Should you agree with these proposed changes, I'd update this PR to fix those.

truckermudgeon:
> how much thought you have put into the eslint config for `maps`

For posterity, the answer is, "some" 🙃. 

I started writing all this code in another repo, without linting. When it came time to create the `maps` repo for publishing, I added linting with stricter defaults, saw a bunch of errors / warnings I didn't care for, dropped down to the default settings which largely matched all my current habits / opinions, then went from there.

I figured if I encountered errors / warnings from settings that didn't match my preferences (like requiring consistent type imports and treating unused vars as warnings instead of errors), I'd override them as needed...

...and it looks like some overriding is needed. So thanks for raising this!

----

> ### 1. Removing properties from objects using destructuring assignments
> I propose to set the [`ignoreRestSiblings: true`](https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings) option on `@typescript-eslint/no-unused-vars`.

✅  SGTM, especially since I've also run into this exact same warning in a wip branch while trying to strip away a property from an object 🙂.

I do have concerns about certain scenarios where the "rest" in a `const { foo, ...rest } = someObj;` destructuring assignment is used, but it's probably best to examine on a case-by-case basis.

> ### 2. Intentionally not using function arguments
> I propose to set the [`argsIgnorePattern: '^_'`](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) option on `@typescript-eslint/no-unused-vars`.

❌ I'd like to leave this as-is for now. I've been meaning to fixup that `LaneControl.tsx` line 😬. It's there because I didn't completely remove some debug code.

> several spots in `maps` where existing code is already written as if the `argsIgnorePattern` option was set.

Hmmmm... I think this is related to the rule behavior where it only warns about unused `_`-prefixed function arguments being unused if they're the last argument:

```ts
function foo(arg0: number, _arg1: number, arg2: number) { // no warning about _arg1
  console.log(arg0, arg2);
}

function foo(arg0: number, arg1: number, _arg2: number) { // warning about _arg2
  console.log(arg0, arg1);
}
```

The (only?) instance where this warning is suppressed is: https://github.com/truckermudgeon/maps/blob/be2be15b98b1cd34ad082b87921b0cac5691126d/packages/clis/generator/geo-json/map.ts#L1181-L1186

where that unused arg is there as a reminder to myself about an idea I wanted to try out. But maybe such reminders are better documented as GitHub issues 🤔 (and then that arg can be deleted, along with a chunk of commented-out logic)


> ### 3. Allowing getters as well as fields to define literals as class properties
> I propose to disable [`@typescript-eslint/class-literal-property-style`](https://typescript-eslint.io/rules/class-literal-property-style/).

⚠️ Interesting... I've never run into this because `maps` doesn't have many classes. I'm guessing there's some code in #47 that could benefit from this change... could you point me to it so I can better understand the use case?

nautofon:
Thanks for your reply!

> For posterity, the answer is, "some" 🙃.

👍

> > I propose to set the [`ignoreRestSiblings: true`](https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings) option on `@typescript-eslint/no-unused-vars`.
> 
> ✅ SGTM, especially since I've also run into this exact same warning in a wip branch while trying to strip away a property from an object 🙂.
>
> I do have concerns about certain scenarios where the "rest" in a `const { foo, ...rest } = someObj;` destructuring assignment is used, but it's probably best to examine on a case-by-case basis.

Right. That may well come up during a review of #47. I use that idiom a few times.  
 

> > I propose to set the [`argsIgnorePattern: '^_'`](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) option on `@typescript-eslint/no-unused-vars`.
> 
> ❌ I'd like to leave this as-is for now.

That's fine. 🙂

> > several spots in `maps` where existing code is already written as if the `argsIgnorePattern` option was set.
> 
> Hmmmm... I think this is related to the rule behavior where it only warns about unused `_`-prefixed function arguments being unused if they're the last argument:

Is that behavior really related to `_`? I think it's just the [`args: 'after-used'`](https://eslint.org/docs/latest/rules/no-unused-vars#args) default.

For what it's worth, I see now that I probably should have proposed to set `args: 'all'` as well. The reason I didn't before was that I didn't understand the rationale behind `args: 'after-used'`. I still don't, but the "being explicit about our intentions" thing doesn't really work without `args: 'all'`.

Here's an example for that rule in action, by the way. The first two params are unused. They're both named with a leading `_`, but that doesn't actually matter to eslint: Currently, with `args: 'after-used'` active, it still doesn't warn even when you remove the `_`.

https://github.com/truckermudgeon/maps/blob/ebc0617d0e946852f2e78d3d29a418be913b83ad/packages/clis/generator/commands/spritesheet.ts#L222-L227

Trying to learn here. Do you by any chance know the reason why `args: 'after-used'` exists (and even is the default)?  
 

> > I propose to disable [`@typescript-eslint/class-literal-property-style`](https://typescript-eslint.io/rules/class-literal-property-style/).
> 
> ⚠️ Interesting... I've never run into this because `maps` doesn't have many classes. I'm guessing there's some code in #47 that could benefit from this change... could you point me to it so I can better understand the use case?

Sure, but perhaps looking at my specific use case—which is [`geographicCoordinatePrecision`](https://github.com/nautofon/maps-nodejs/blob/216c0a85aeb333d9e41196255412a0d914abf977/packages/clis/generator/geo-json/extra-labels.ts#L700-L703)—is better left for a review of #47. FWIW, my goal there was to offer a way to change a constant on a per-instance basis. Not only are there other ways to do that, but you could even argue [YAGNI](https://en.wikipedia.org/wiki/YAGNI).

The point is that in my view, the eslint error is just *wrong*, irrespective of #47. The description of `class-literal-property-style` makes it sound like this is merely a stylistic choice, and that it's only about consistency. It isn't.

Let me try to give a slightly more complete example of that:

```ts
class A {
  foo = 'A.foo';
  get bar() { return 'A.bar' }

  constructor () {
    console.log(this.foo + ' / ' + this.bar);
  }
}

class B extends A {
  foo = 'B.foo';
  get bar() { return 'B.bar' }
}

new B();
// output: A.foo / B.bar
```
You might expect `B.foo` to override `A.foo`. But that only happens *after* the constructor of `A` has run.

There *is* a functional difference between getters and fields. Therefore, trying to prohibit one or the other "for a consistent style" (which is the reasoning behind this eslint rule) seems to be inappropriate.

(Or am I missing something?)

truckermudgeon:
> Is that behavior really related to `_`? I think it's just the [`args: 'after-used'`](https://eslint.org/docs/latest/rules/no-unused-vars#args) default.

Ahhhh... I have learned some things. Thanks for the discussion!

You're right. The warning for the code I quoted has nothing to do with the fact that the arg is prefixed with `_`.

Also:

```ts
// ESLint gives no warning 
/*eslint no-unused-vars: ["warn", { "args": "after-used" }]*/
mask.scan(0, 0, mask.bitmap.width, mask.bitmap.height, (x, y, i) => {


// ESLint warns 
/*eslint no-unused-vars: ["warn", { "args": "all" }]*/
mask.scan(0, 0, mask.bitmap.width, mask.bitmap.height, (x, y, i) => {
```

And I agree: if this rule were changed, then it'd make sense to set the ignore pattern _and_ use `args: 'all'`.

In `maps`, turns out there's a bit of a hybrid behavior happening, because of [this compiler option](https://github.com/truckermudgeon/maps/blob/main/tsconfig.base.json#L13).

The current lint settings don't warn for the `(x, y, i)` callback, but the current compiler options will cause TypeScript compilation to fail:

```
packages/clis/generator/commands/spritesheet.ts:223:59 - error TS6133: 'x' is declared but its value is never read.

223   mask.scan(0, 0, mask.bitmap.width, mask.bitmap.height, (x, y, i) => {
                                                              ~

packages/clis/generator/commands/spritesheet.ts:223:62 - error TS6133: 'y' is declared but its value is never read.

223   mask.scan(0, 0, mask.bitmap.width, mask.bitmap.height, (x, y, i) => {
```

Also, TypeScript seems to have a built-in `^_` ignore pattern set, which is why the current code with `_` prefixes compiles without error.

So... there's a mix of where enforcement is happening, which isn't obvious (I even forgot about it 😳).

Since the compiler behaves as if `argsIgnorePattern: '^_'` were set, then: we might as well mirror that in the linter settings, too 🙂 (and also set `args: 'all'`).

(And: part of me wants to convert all those compiler options into the equivalent linter options, so there's one source of truth for where these errors/warnings come from. But another part of me is lazy, and also trusts the TS compiler more in its ability to catch lint-y things quicker than ESLint can.)

> Do you by any chance know the reason why `args: 'after-used'` exists (and even is the default)?

I can only guess that it's used for cases like the spritesheet callback, where we're only interested in some non-first, non-last arg(s)? E.g.:

```ts
// args: after-used
const a = [1,2,3];
a.forEach((value, index, array) => {
  console.log(index);
});
// doesn't warn on value, which is desired
// warns on array, which is desired (fix by not declaring the array arg)
```

```ts
// args: all
const a = [1,2,3];
a.forEach((value, index, array) => {
  console.log(index);
});
// warns on value, which is undesired
// warns on array, which is desired (fix by not declaring the array arg)
```


Comments / musings about `class-literal-property-style` to come later 😅

nautofon:
> In `maps`, turns out there's a bit of a hybrid behavior happening, because of [this compiler option](https://github.com/truckermudgeon/maps/blob/main/tsconfig.base.json#L13).
> 
> The current lint settings don't warn for the `(x, y, i)` callback, but the current compiler options will cause TypeScript compilation to fail:

Oooh I see! Good catch. Pretty sure I got these compiler errors some time earlier, but misattributed them to eslint in my head.

> Also, TypeScript seems to have a built-in `^_` ignore pattern set, which is why the current code with `_` prefixes compiles without error.

Right. The only place I see that being documented is the [TS 2.0 changelog](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals), which is easy to miss.

> Since the compiler behaves as if `argsIgnorePattern: '^_'` were set, then: we might as well mirror that in the linter settings, too 🙂 (and also set `args: 'all'`).

All right, I've put it back into this branch and added `args: 'all'`, too.

> ```ts
> a.forEach((value, index, array) => {
>   ...
> // warns on array, which is desired (fix by not declaring the array arg)
> ```
Ahyes, of course. 🤦 It seems I forgot that JS doesn't do strict arity checks, and that you can just shorten the argument list like that. That would explain `after-used`.

Thank you for your detailed explanations! I have learned some things, too. ☺️

truckermudgeon:
> The description of `class-literal-property-style` makes it sound like this is merely a stylistic choice, and that it's only about consistency. It isn't.

That's fair to say: getters that return literals _do_ behave differently from literal fields.

I'd put this particular linter option in the "I wouldn't expect to run into this warning, because I wouldn't write code that triggers it" bucket.

If I were writing a base class and an extending class, where the base class declares something (either a field or a getter) that represents _any_ constant value (whether it's a literal like the number `42`, or the immutable array produced by `Object.freeze([4,2])`), then I'd do that with a `readonly field`, because a getter seems unnecessary (the return value doesn't change, so no getter code needs to be evaluated).

Assuming that `foo` isn't meant to change during the lifetime of an `A` instance, or during the lifetime of an instance of a class derived from `A`, and that its value should be set "at the beginning" of its life, then I'd expect that field to be marked as `readonly`.

> ```ts
> class A {
>   readonly foo = 'A.foo';
> ```

Adding `readonly` will cause the `foo` declaration in `B` in the examnple to throw a compiler error, but I'd take that as a sign that maybe there's a better way to express the data constraints whilst still getting the desired behavior, while also sidestepping the question of "what value will `this.foo` produce when referenced in a base-class constructor?"

Also: I'd expect getters to be used for computed values that can change during an object's lifetime, only. Maybe that's the style-related consistency that the rule is meant to enforce? IMO, it seems like using a getter to return a constant value is useful only for working around quirks of the JavaScript language (like the fact that a base class ctor doesn't have access to the value of a derived class' field override, or to hide that field from being included in a `const { ...fields } = this` destructuring, etc).


One more thing I learned while digging into this problem: I tried the equivalent of the A/B example in Java: 

image

Java doesn't have getters like JS/TS, so expectations about the behavior you'd get are a bit more explicit because you see when you're making a function call vs. accessing a field... but the behavior that a base ctor doesn't have access to the value of a derived field matches JS/TS.

Maybe this is a common behavior in OOP languages (and OOP-ish languages like JS)?

nautofon:
> I'd put this particular linter option in the "I wouldn't expect to run into this warning, because I wouldn't write code that triggers it" bucket.
> [...]

Right! That's actually an important observation, as I'll get to in a moment.

> I'd expect getters to be used for computed values that can change during an object's lifetime, only. Maybe that's the style-related consistency that the rule is meant to enforce?

I don't think so.

> IMO, it seems like using a getter to return a constant value is useful only for working around quirks of the JavaScript language (like the fact that a base class ctor doesn't have access to the value of a derived class' field override, or to hide that field from being included in a `const { ...fields } = this` destructuring, etc).

My point exactly: Both constant fields and constant getters *can* be useful, in specific circumstances.

In the abstract, I don't see a reason to prohibit one of them outright, as this eslint rule is trying to do.

But now I looked at the history for the rule and found [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/1582#issuecomment-585364060) by its original author. He essentially says that fields are better for apps, but getters are better for libraries. That makes a lot of sense: I *have* been approaching #47 with a library mindset; OTOH for you as the author of `maps`, I think it's only natural *not* to do that. That might be the reason why you don't expect to write code that triggers this rule.

I still feel the rule probably isn't helping `maps`. But when I'm only thinking about the app, then YAGNI is really the only valid answer to my specific use case. So the rule might be a non-issue. Hm. Interesting.

> One more thing I learned while digging into this problem: I tried the equivalent of the A/B example in Java:
> [...]
> Maybe this is a common behavior in OOP languages (and OOP-ish languages like JS)?

No, it differs between object systems.

Many object systems don't allow overwriting fields in the first place, for instance C++ or Java. In your Java example, `A.foo` and `B.foo` are actually two entirely different fields. In JS however, `A.foo` and `B.foo` are one and the same field. So the kind of situation we're talking about with JS cannot occur in an object system like Java's. It isn't really possible to compare.

Not sure how common JavaScript's behavior of setting field values in-between running constructors is. It's the only system I know with this particular quirk. 

One example for an object system that *does* allow overwriting fields, but does *not* share JavaScript's quirk, would be Moose (for Perl). Moose sets sub-class field values before running the constructors.

truckermudgeon:
> He essentially says that fields are better for apps, but getters are better for libraries. That makes a lot of sense

👍 

> I still feel the rule probably isn't helping `maps`. But when I'm only thinking about the app, then YAGNI is really the only valid answer to my specific use case. So the rule might be a non-issue. Hm. Interesting.

I think it's a non-issue for now. If the packages under `libs` (which, currently, are pretty trivial, not well-organized, and not really thought out) ever export classes or instances of classes, then that would be the time to re-evaluate the rule and consider adding library-y eslint overrides that target `libs/*`.

> No, it differs between object systems.
>
> [...]

Ahhh... TIL... except for this part:

> [JS] is the only system I know with this particular quirk [of setting field values in-between running constructors]

If there's anything I've learned in my time with JS, it's that it's a quirky language 🙃

nautofon:
> If there's anything I've learned in my time with JS, it's that it's a quirky language 🙃

No argument there. 😄

Okay, I removed the `class-literal-property-style` change from this patch.

truckermudgeon (approved):

PR #51, opened 2025-02-16
Make parser capable of reading ETS2 hex-escaped Unicode strings

nautofon:
The ETS2 demo's `mileage_targets.sii` file contains strings with UTF-8 bytes escaped with `\x`. This patch makes the parser capable to decode those to regular characters.

The demo version is actually somewhat outdated, so I can't know if this patch is necessary to correctly read the files from the full version. But I suspect that it is. You would have to check `europe-mileageTargets.json`.


truckermudgeon (approved):
> You would have to check `europe-mileageTargets.json`.

Yep, looks like hex-escaped unicode chars are present there. Thanks for the fix!

PR #58, opened 2025-07-21
demo,navigator,ui: add tileRootUrl prop to map components

truckermudgeon:
This PR:
- updates map components with a `tileRootUrl` prop, that controls where `.pmtiles` are fetched from
- updates the `demo` app to use a `tileRootUrl` based on an [env variable](https://vite.dev/guide/env-and-mode)
- updates the `navigator` app to use `''` for its `tileRootUrl`

Note that the changes in this PR do not lead to any behavioral changes. Because of:
- the `.env` file that's checked into the `demo` app package, and
- the hardcoded `''` `tileRootUrl` used by the `navigator` app package

development- and production-mode builds of `demo` and `navigator` will still search for `.pmtiles` files at the root of whatever server is hosting the webapps.

The only meaningful change happens locally on my MacBook: I have a `packages/apps/demo/.env.production.local` file that looks like:

```
VITE_TILE_ROOT_URL=https://maps.truckermudgeon.com
```

and it ensures that my production-mode builds of `demo` (which I upload to GitHub Pages) will fetch `.pmtiles` from Cloudflare instead of from GitHub Pages. I'm doing this so that Firefox users can visit https://truckermudgeon.github.io without [any problems related to GitHub Pages](https://github.com/truckermudgeon/truckermudgeon.github.io/issues/4).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ npm-debug.log*
 tsconfig.tsbuildinfo
 .eslintcache
 *storybook.log
+*.local
--- a/packages/apps/demo/.env
+++ b/packages/apps/demo/.env
@@ -1,0 +1,3 @@
+# root url where .pmtiles are stored, e.g., https://truckermudgeon.github.io
+# leave blank to use the host URL of the demo app's webserver.
+VITE_TILE_ROOT_URL=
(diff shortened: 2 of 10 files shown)


coyote74mods:
Dear truckermudgeon I know this is asking a lot, could you give me a guide on how to do it? I'd love to use the map, but at 60, I'm lost on how to do it. I'd be very grateful.

PR #62, opened 2025-07-30
generator,demo: add company searching

truckermudgeon:
This PR:
- adds a `generator companies` command to generate GeoJSON for companies
- updates `demo` to consume this GeoJSON, and present a simple UI for searching companies, similar to the UI for searching achievements

image

<br>
<br>

Thanks to [Hillycamp](https://forum.scssoft.com/viewtopic.php?p=2021669&sid=9f2542a483c02393696147af3f78572f#p2021669) for the idea!

A future PR may simplify the search UI so that searching for cities and companies (and _maybe_ achievements) will be combined into a single search bar, instead of having dedicated search bars for each 🤔


packages/apps/demo/package.json
```diff
@@ -24,7 +24,7 @@
     "@emotion/react": "^11.11.3",
     "@emotion/styled": "^11.11.0",
     "@fontsource/inter": "^5.0.16",
-    "@mui/icons-material": "^5.15.13",
+    "@mui/icons-material": "^7.2.0",
```
truckermudgeon, line 27: Updates just so I could get the [`Pallet`](https://mui.com/material-ui/material-icons/?query=pallet) icon 😬

packages/apps/demo/src/OmniBar.tsx
```diff
@@ -161,38 +162,131 @@ export const OmniBar = (props: OmniBarProps) => {
     onAchievementSelect(achievementOption, { enableFitBounds: false });
   }, [achievementOption, props.visibleStates]);
 
+  const [companyOption, setCompanyOption] = useState<CompanyOption | null>(
+    null,
+  );
+  const onCompanySelect = React.useCallback(
```
truckermudgeon: Pretty much a copy/paste of `onAchievementSelect` above... which looks to have a minor callback-deps bug.

Will fix and unify.

packages/apps/demo/src/OmniBar.tsx
```diff
@@ -161,38 +162,131 @@ export const OmniBar = (props: OmniBarProps) => {
...
-        )}
+        <SearchBar
+          selected={gameMap.value}
+          onMapSelect={onMapSelect}
+          onCitySelect={onCitySelect}
+          onCompanySelect={onCompanySelect}
+          onAchievementSelect={onAchievementSelect}
+          visibleStates={props.visibleStates}
+          visibleStateDlcs={props.visibleStateDlcs}
+        />
```
truckermudgeon, line 199: Better to pass in Components instead of callbacks?

packages/clis/generator/commands/companies.ts
```diff
@@ -0,0 +1,163 @@
...
+import { createIsoA2Map } from '../geo-json/populated-places';
+import { logger } from '../logger';
+import type { MappedDataForKeys } from '../mapped-data';
+import { readMapData } from '../mapped-data';
+import { writeGeojsonFile } from '../write-geojson-file';
+import { maybeEnsureOutputDir, untildify } from './path-helpers';
+
+export const command = 'companies';
+export const describe =
+  'Generates companies.geojson from map-parser JSON files';
```
truckermudgeon, line 20: Need to update the README and Makefile to reference this new command.

packages/clis/generator/write-geojson-file.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'fs';
 
 export function writeGeojsonFile(
   path: string,
-  data: GeoJSON.FeatureCollection,
+  data: GeoJSON.FeatureCollection & { properties?: Record<string, unknown> },
```
truckermudgeon, line 5: The spec says a FeatureCollection can have [foreign members](https://datatracker.ietf.org/doc/html/draft-ietf-geojson-03#section-6)... so I'm allowing callers to provide a `properties` object to store arbitrary data (like token-to-label lookup tables).

packages/clis/generator/commands/companies.ts
```diff
@@ -0,0 +1,163 @@
...
+  company: CompanyItem,
+  context: MappedDataForKeys<['cities', 'countries', 'prefabs']> & {
+    companyQuadtree: Quadtree<QtCompanyEntry>;
+    mapCountryCode: (countryCode: string) => string;
+  },
+): CompanyFeature {
+  const poi = assertExists(context.companyQuadtree.find(company.x, company.y));
+  if (distance(poi, company) > 50) {
+    logger.warn(
+      'poi is > 20 meters away from company item',
```
truckermudgeon: Warning message and actual check don't match 😬

packages/clis/generator/commands/companies.ts
```diff
@@ -0,0 +1,163 @@
...
+      `0x${company.uid.toString(16)}`,
+    );
+  }
+
+  const prefab = assertExists(context.prefabs.get(company.prefabUid));
+  const city = assertExists(context.cities.get(company.cityToken));
+  const country = assertExists(context.countries.get(city.countryToken));
+
+  return {
```
truckermudgeon: Prefer `turf.point` helper (I've been starting to use them more).

packages/apps/demo/src/OmniBar.tsx
truckermudgeon: Will worry about this later. May be irrelevant with a truly omni omni-searchbar.

packages/clis/generator/commands/companies.ts
truckermudgeon: https://github.com/truckermudgeon/maps/pull/62/commits/4b77f3968051290bdb09bfa6437818b3a5cfc382

packages/clis/generator/commands/companies.ts
truckermudgeon: https://github.com/truckermudgeon/maps/pull/62/commits/ce1904774e0be35d48c448a2e25890b051b71aae

packages/clis/generator/commands/companies.ts
truckermudgeon: https://github.com/truckermudgeon/maps/pull/62/commits/38b8a95fcc6903f132bf19038ebcf46403259d14

packages/apps/demo/src/OmniBar.tsx
truckermudgeon: https://github.com/truckermudgeon/maps/pull/62/commits/67379ab840f53dde89ef848738c4ad71917e1775

packages/clis/generator/write-geojson-file.ts
nautofon: Technically the spec says FeatureCollection objects MUST NOT contain a `properties` member. Only member names not described in the spec are considered "foreign".

https://datatracker.ietf.org/doc/html/rfc7946#section-7.1

Giving it a different name (say, `lookupTable` instead of `properties`) would make it conforming.

truckermudgeon: TIL... I should've read the spec closer 🫠 

All the code in this PR is going to be deleted soon because of the new search stuff, so I'll be removing the `& { properties?: Record<string, unknown> }` typing... but I'll keep the spec in mind if/when the need for adding top-level metadata comes up again.

PR #63, opened 2025-08-13
demo: add context menu

truckermudgeon:
This PR adds a context menu to the demo app. Currently, it: 

* shows game coordinate _and_ long/lat readouts, as per my interpretation of "Coordinate readout (game coords? long/lat?)" from @nautofon's [ideas list](https://gist.github.com/nautofon/d6b3fb841f632478c6db6f0d7f00231e):
  <br>image<br><br>
* provides a Google-Maps-ish UI to draw a linestring onto the map to measure distances:
  <br>image

Future PRs will add menu items to route to/from the right-clicked point.

### DIFF
--- a/packages/apps/demo/src/ContextMenu.tsx
+++ b/packages/apps/demo/src/ContextMenu.tsx
@@ -14,8 +14,14 @@ import {
 } from '@mui/joy';
 import type { VirtualElement } from '@popperjs/core/lib/types';
 import { assert, assertExists } from '@truckermudgeon/base/assert';
+import type { Extent } from '@truckermudgeon/base/geom';
+import { center, withinExtent } from '@truckermudgeon/base/geom';
 import { UnreachableError } from '@truckermudgeon/base/precon';
-import { fromWgs84ToAtsCoords } from '@truckermudgeon/map/projections';
+import {
+  fromWgs84ToAtsCoords,
+  fromWgs84ToEts2Coords,
+} from '@truckermudgeon/map/projections';
+import { distance } from '@turf/distance';
 import type { GeoJSON } from 'geojson';
 import type {
   GeoJSONSource,
@@ -27,23 +33,28 @@ import { useMap } from 'react-map-gl/maplibre';
 
 interface ClickContext {
   anchorEl: VirtualElement;
-  position?: {
+  closestGame: 'ats' | 'ets2';
+  position: {
     lngLat: [number, number];
-    xz: [number, number];
+    // `xz` is undefined when context menu is brought up at a point that
+    // is not in bounds of the ATS or ETS2 map.
+    xz: [number, number] | undefined;
   };
 }
 
-const toFixed = (f: number, tuple: [number, number]): [number, number] =>
-  tuple.map(v => Number(v.toFixed(f))) as [number, number];
+type PointFeature = GeoJSON.Feature<GeoJSON.Point, { id: number }>;
 
-let idCounter = 0;
-const newId = () => {
-  const stringId = idCounter.toString(36);
-  idCounter++;
-  return stringId;
+// TODO read these values from the .pmtiles files at runtime.
+const extents = {
+  ats: [
+    [-124.477162, 25.767968].map(n => Math.floor(n)),
+    [-88.777474, 49.1223839].map(n => Math.ceil(n)),
+  ].flat() as Extent,
+  ets2: [
+    [-10.025698, 34.897275].map(n => Math.floor(n)),
+    [33.284941, 61.881437].map(n => Math.ceil(n)),
+  ].flat() as Extent,
 };
-
-type PointFeature = GeoJSON.Feature<GeoJSON.Point, { id: string }>;
 
 export const ContextMenu = () => {
   console.log('render ContextMenu');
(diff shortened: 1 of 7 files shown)


packages/libs/map/projections.ts
```diff
@@ -75,3 +83,33 @@ export const fromEts2CoordsToWgs84 = ([x, y]: Position): Position => {
...
+    unUkX * ukScaleFactor < calais[1];
+  if (isUk) {
+    x = unUkX;
+    y = unUkY;
+  }
+
+  return [
+    x + ets2DefData.mapOffset[0], //
+    y + ets2DefData.mapOffset[1],
+  ].map(v => Number(v.toPrecision(10))) as [number, number];
```
truckermudgeon: Using `toPrecision` as a hacky way to round, and prevent test failures like:

image

10 sigfigs is probably overkill, and on second thought, it isn't what I originally intended. Should be `toFixed(10)`.

packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -0,0 +1,551 @@
...
+    lngLat: [number, number];
+    // `xz` is undefined when context menu is brought up at a point that
+    // is not in bounds of the ATS or ETS2 map.
+    xz: [number, number] | undefined;
+  };
+}
+
+type PointFeature = GeoJSON.Feature<GeoJSON.Point, { id: number }>;
+
+// TODO read these values from the .pmtiles files at runtime.
```
truckermudgeon, line 51: These values come from drag/dropping `{ats,ets2}.pmtiles` files onto https://pmtiles.io/archive/

image

packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -0,0 +1,551 @@
...
+    [-124.477162, 25.767968].map(n => Math.floor(n)),
+    [-88.777474, 49.1223839].map(n => Math.ceil(n)),
+  ].flat() as Extent,
+  ets2: [
+    [-10.025698, 34.897275].map(n => Math.floor(n)),
+    [33.284941, 61.881437].map(n => Math.ceil(n)),
+  ].flat() as Extent,
+};
+
+export const ContextMenu = () => {
```
truckermudgeon, line 63: There's a lot going on here 😬. So much that I'm kinda just glazing over the walls of code.

Will break this up in the future, e.g., when adding "route to/from here" menu items to the context menu.

packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -0,0 +1,551 @@
...
+      </div>
+      <MeasuringToast
+        measuring={measuring}
+        close={closeMeasuringToast}
+        measuringPoints={measuringPoints}
+      />
+      <CopiedToClipboardToast
+        open={showClipboardToast}
+        close={closeClipboardToast}
+      />
```
truckermudgeon, line 367: Having a context menu component also be responsible for showing + managing the UI triggered by the context menu items feels a little 🤢. But I'm too lazy to do anything about it right now.

packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -0,0 +1,551 @@
...
+        case 'lngLat':
+          void navigator.clipboard.writeText(
+            pos.lngLat.map(n => toFixed(n, 4)).join(','),
+          );
+          break;
+        case 'xz':
+          void navigator.clipboard.writeText(
+            assertExists(pos.xz)
+              .map(n => toFixed(n, 1))
+              .join(';'),
```
nautofon, line 97: JFTR, I would personally prefer `,` as a separator for game coordinates on the clipboard. That way, it'd be easier to copy-paste them into CSV.

nautofon (commented):
Hey, very cool! When I wrote that, I wasn't quite sure if both types of coordinates would be actually useful, hence the question mark. But I really like how you show both next to each other here.

It won't surprise you to hear me say that I'd prefer "east" and "south" as labels for the game coordinates. 🙂 That way, it'd be immediately clear what each number refers to.

Should you decide to stick with "x" and "z", I think you should at least make sure that geographical and game coordinates use the same axis order. That's currently not the case: Geographical coords are ordered `lat / lng` (north-south axis first), game coords are ordered `x / z` (east-west axis first). Swapping the order on *one* of these can help users logically deduce what the somewhat cryptic labels "x" and "z" stand for.

In other words, any one of the following would be a minor improvement (listed in my order of preference):

* `lat / lng` + `east / south`
* `lng / lat` + `x / z`
* `lat / lng` + `z / x`

truckermudgeon:
> Should you decide to stick with "x" and "z", I think you should at least make sure that geographical and game coordinates use the same axis order. 

I did consider this... and it did bug me that the axis orders were mixed. I ended up going with the mixed orders because of the conventions in which they're used: lng / lat because the only other user-facing geo coords in the app are in the URL hash, which uses lng / lat. And other webmaps and QGIS use lng / lat, too (QGIS is interesting: when I'm using it to debug my GeoJSON data, the coordinates readout in the status bar uses lng / lat when I'm using the EPSG:4326 CRS, but it'll use easting/southing when using my user-defined CRS for ATS).

🤔 maybe the labels for the coordinates are enough to signal what's going on, because I personally would prefer `lng / lat` + `x / z`.

As for `,` vs `;` for game coords: I figured it'd be nice to match how they appear in SCS bug reports, which use semicolons. But maybe that doesn't really apply, because I'm not including a `y` value, and so they can't be copied as-is, anyway.

Thanks for the feedback. Expect changes to the ordering and game-coord separator, soon 🙂

nautofon:
> 🤔 maybe the labels for the coordinates are enough to signal what's going on, because I personally would prefer `lng / lat` + `x / z`.

Yes, I'd support changing it to `lng / lat`. (You currently use `lat / lng`.)

> (QGIS is interesting: when I'm using it to debug my GeoJSON data, the coordinates readout in the status bar uses lng / lat when I'm using the EPSG:4326 CRS, ...

That's kinda weird. EPSG:4326 is supposed to be `lat / lng`. OGC:CRS84, on the other hand, is `lng / lat`.

I think the order doesn't matter much as long as both axes are clearly labeled.

PR #64, opened 2025-08-13
demo: add street view proof-of-concept

truckermudgeon:
This PR adds a [`/street-view` page](https://truckermudgeon.github.io/street-view) to the demo app:

image

Some discussion about the street view demo can be found in [this thread](https://forum.scssoft.com/viewtopic.php?p=2053860#p2053860) of the SCS Forums.

The `generator` code that generates screenshotting and panorama-creation scripts will come in a follow-up PR.

### DIFF
--- a/packages/apps/demo/src/StreetViewDemo.tsx
+++ b/packages/apps/demo/src/StreetViewDemo.tsx
@@ -1,0 +1,230 @@
+import { EquirectangularTilesAdapter } from '@photo-sphere-viewer/equirectangular-tiles-adapter';
+import type { MarkerConfig } from '@photo-sphere-viewer/markers-plugin';
+import { MarkersPlugin } from '@photo-sphere-viewer/markers-plugin';
+import '@photo-sphere-viewer/markers-plugin/index.css';
+import type {
+  GpsPosition,
+  VirtualTourNode,
+  VirtualTourPluginConfig,
+} from '@photo-sphere-viewer/virtual-tour-plugin';
+import { VirtualTourPlugin } from '@photo-sphere-viewer/virtual-tour-plugin';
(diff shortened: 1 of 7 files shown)


packages/apps/demo/src/StreetViewDemo.tsx
```diff
@@ -0,0 +1,230 @@
...
+          tileRootUrl={props.tileRootUrl}
+          mode={'light'}
+        ></BaseMapStyle>
+        <GameMapStyle
+          tileRootUrl={props.tileRootUrl}
+          game={'ats'}
+          mode={'light'}
+          enableIconAutoHide={true}
+          visibleIcons={allIcons}
+          dlcs={AtsSelectableDlcs}
```
truckermudgeon, line 223: Reminder to self: if/when street view and/or photo spheres are added to the demo app, then the data for panoramas will need to be tagged with dlc guards 🤔

packages/apps/demo/src/StreetViewDemo.tsx
```diff
@@ -0,0 +1,230 @@
...
+      nodeId: `node-${index}`,
+    }));
+
+  const tourConfig: VirtualTourPluginConfig = {
+    positionMode: 'gps',
+    nodes: [
+      { ...node(0), links: links(1) },
+      { ...node(1), links: links(0, 2) },
+      { ...node(2), links: links(1, 3) },
+      { ...node(3), links: links(2, 8) },
```
truckermudgeon, line 97: Auto-generating links for connected street-view panos will be interesting to solve... e.g., if panos are only taken at nav-graph nodes, then we could generate a bunch of routes from every node in the graph to some fixed point, in both forward/backward directions. Then we'd have a picture of how all the nodes are connected 🤔.

PR #67, opened 2025-08-25
generator,ui: add exit numbers to map

truckermudgeon:
This PR:
- updates the `generator map` command to detect and emit ExitFeatures, using the [sign data](https://github.com/truckermudgeon/maps/commit/69071108059c6f0d6380c3a798ec046a58c84b21) now being produced by the parser.
- updates the `ui/GameMapStyle` component to render ExitFeatures

image

<br>
<br>

ExitFeatures are detected with heuristics that _seem_ to produce decent results on the ATS map. If any readers spot an exit sign on the map that shouldn't be there, please right-click on it to get its coordinates and raise an issue 🙂


packages/apps/demo/src/Demo.tsx
```diff
@@ -157,6 +158,18 @@ const Demo = (props: { tileRootUrl: string; pixelRootUrl: string }) => {
     showPhotoSpheresUi,
   ]);
 
+  const onLoad = async () => {
```
truckermudgeon: Is this the best place to do this? It'd be nice if `GameMapStyle` handled it somehow.

packages/clis/generator/mapped-data.ts
```diff
@@ -167,18 +167,20 @@ export function readMapData<
 
   const focusXY =
     focusCoords != null
-      ? (i: { x: number; y: number }, padding = 0) =>
```
truckermudgeon, line 170: The `signs` array is large: about 72k for ATS. During debugging, when focused on a city like Las Vegas, I was running into assertion errors because a sign item tens of thousands of meters away from Las Vegas was unexpectedly included in GeoJSON generation. Turned out to be a rookie mistake on my part 🤦.

`focusXY` is often called like:

```ts
const filteredItems = someArray.filter(focusXY);
```

which is the equivalent of:

```ts
// padding increases as index increases. items far away from the focus point end up included.
const filteredItems = someArray.filter((item, index) = focusXY(item, index)); 
```

But the intent is usually:

```ts
// padding isn't supplied, so it should default to 0.
const filteredItems = someArray.filter((item, index) = focusXY(item));
```

I've removed the concept of `padding` from `focusXY` altogether. If you want padding, use `focusXYPlus`.

Issue #70, opened 2025-09-02
Implementing a Ruler to measure distances


NiNo-3:
I'd like for there to be a ruler to measure distances between two points, since I want to build a Minecraft map and your maps seem like the best reference.

truckermudgeon:
Would the "Measure distance" right-click menu option (see https://github.com/truckermudgeon/maps/pull/63) do what you want?

NiNo-3:
Oh, I can't access it it seems, as I only have a mere phone.

truckermudgeon:
Ah, that's on me... I'll see if I can trigger the context menu with a long-press. Might take me a while, though, as it's a busy time for me at the moment.

I'll update this issue when I have something 🙂

NiNo-3:
Okay!

NiNo-3:
Also is it possible to add a satellite map or?

NiNo-3:
> Ah, that's on me... I'll see if I can trigger the context menu with a long-press. Might take me a while, though, as it's a busy time for me at the moment.
> 
> I'll update this issue when I have something 🙂

No worries!

truckermudgeon:
Just pushed a fix... I only tested using browser tools and not an actual mobile device 😅. Let me know if it fixes things for you 🙂 

> Also is it possible to add a satellite map or?

This is _possible_, but it'd be way too much work (gotta do a ton of in-game screenshots), and it'd look pretty crappy: there's a lot of weirdness in the map that's only revealed with birdseye views, like holes in the map, extremely skinny mountain ranges because they only need to be viewed from one angle, etc.

NiNo-3:
Yeah it works

NiNo-3:
> This is _possible_, but it'd be way too much work (gotta do a ton of in-game screenshots), and it'd look pretty crappy: there's a lot of weirdness in the map that's only revealed with birdseye views, like holes in the map, extremely skinny mountain ranges because they only need to be viewed from one angle, etc.
 
No problem, it's just that I wanted to build the map in Minecraft and I needed the satellite to see the road types or something, but it doesn't matter.

NiNo-3:
Also where did you get the building data?

PR #71, opened 2025-09-05
demo: trigger context menu on longpress

truckermudgeon:
See https://github.com/truckermudgeon/maps/issues/70

The context menu added in #63 relies on the `contextmenu` event to be fired... which isn't always a thing on mobile browsers (e.g., mobile Safari doesn't support the `contextmenu` event).

This PR hacks a longpress event handler so that mobile users can summon the context menu by tapping-and-holding for one second:

https://github.com/user-attachments/assets/cb8a1008-0322-4e0d-a921-6de09d8f57f8

### DIFF
--- a/packages/apps/demo/src/ContextMenu.tsx
+++ b/packages/apps/demo/src/ContextMenu.tsx
@@ -110,7 +110,7 @@ export const ContextMenu = () => {
 
   useEffect(() => {
     const showContextMenu = (e: MapLayerMouseEvent) => {
-      const { clientX, clientY } = e.originalEvent;
+      const { x, y } = e.point;
       const lngLat = e.lngLat.toArray();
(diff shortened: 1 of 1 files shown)


packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -422,7 +441,7 @@ const MeasuringToast = memo(
                 Click on the map to add to your path. Click on an existing point
                 to remove it from the path.
               </Typography>
-              <Stack direction={'row'} gap={2} sx={{ textWrap: 'nowrap' }}>
+              <Stack direction={{ xs: 'column', md: 'row' }} gap={2}>
```
truckermudgeon, line 444: Before this change, the measuring toast would get cut off in portrait orientation:

image

PR #72, opened 2025-09-13
generator: add search command

truckermudgeon:
This PR adds a `generator search` command to generate GeoJSON for almost all labeled POIs: cities, scenery towns, companies, viewpoints, ferry terminals / train stations, photo trophies, and truck dealers.

This GeoJSON will be consumed in a follow-up PR to update the `demo` app's search UI, so that it can look something like:

image

where:
- there are only two search modes: one for POIs, and one for Acheivements
- users can enter search queries like `las vegas companies` or `peterbilt dealer` to see a list of relevant results

### DIFF
--- a/packages/clis/generator/commands/search.ts
+++ b/packages/clis/generator/commands/search.ts
@@ -53,7 +53,8 @@ export const builder = (yargs: Argv) =>
     })
     .option('extraLabels', {
       alias: 'x',
-      describe: 'Path to extra-labels.geojson file (required for usa)',
+      describe:
+        'Path to extra-labels.geojson file (required for usa, ignored for europe)',
       type: 'string',
       coerce: untildify,
       demandOption: false,
@@ -164,11 +165,9 @@ export function handler(args: BuilderArguments<typeof builder>) {
       throw new UnreachableError(args.map);
   }
 
-  const spatialIndices = createSpatialIndices(tsMapData, sceneryTowns);
-
   const context = {
     ...tsMapData,
-    ...spatialIndices,
+    ...createSpatialIndices(tsMapData, sceneryTowns),
     dlcGuardQuadTree,
   };
 
@@ -194,22 +193,19 @@ export function handler(args: BuilderArguments<typeof builder>) {
         'unknown country code: ' + f.properties.country,
       );
     }
-    return {
-      ...f,
-      properties: {
-        dlcGuard: assertExists(
-          context.dlcGuardQuadTree.find(
-            f.geometry.coordinates[0],
-            f.geometry.coordinates[1],
-          ),
-        ).dlcGuard,
-        stateName: state.name,
-        stateCode: state.code,
-        label: f.properties.text,
-        tags: ['scenery'],
-        type: 'scenery',
-      },
-    };
+    return point(f.geometry.coordinates, {
+      dlcGuard: assertExists(
+        context.dlcGuardQuadTree.find(
+          f.geometry.coordinates[0],
+          f.geometry.coordinates[1],
+        ),
+      ).dlcGuard,
+      stateName: state.name,
+      stateCode: state.code,
+      label: f.properties.text,
+      tags: ['scenery'],
+      type: 'scenery',
+    });
   });
 
   const normalizeCoords = createNormalizeFeature(args.map, 4);
@@ -354,10 +350,6 @@ function poiToSearchFeature(
   },
 ): SearchFeature[] {
   const { dlcGuardQuadTree, nodeQuadTree, cityQuadTree, cityRTree } = context;
-  const geometry: GeoJSON.Point = {
-    type: 'Point',
-    coordinates: [poi.x, poi.y],
-  };
   const getDlcGuard = (p: { x: number; y: number }): number =>
     assertExists(dlcGuardQuadTree.find(p.x, p.y)).dlcGuard;
 
@@ -386,6 +378,8 @@ function poiToSearchFeature(
       maxX: poi.x,
       maxY: poi.y,
     })
+    // use `.at(0)` instead of `[0]` to force inferred type of `containingCity`
+    // to be `| undefined`.
     .at(0);
   const nearestCity = assertExists(cityQuadTree.find(poi.x, poi.y));
   const cityProperties: { containingCity?: string; nearestCity?: string } = {
@@ -518,13 +512,8 @@ function poiToSearchFeature(
       'as nearest city.',
     );
   }
-  return [
-    {
-      type: 'Feature',
-      properties,
-      geometry,
-    },
-  ];
+
+  return [point([poi.x, poi.y], properties)];
 }
 
 function cityToSearchFeature(
@@ -539,25 +528,17 @@ function cityToSearchFeature(
     city.x + cityArea.width / 2,
     city.y + cityArea.height / 2,
   ];
-  const geometry: GeoJSON.Point = {
-    type: 'Point',
-    coordinates,
-  };
-  const getDlcGuard = (p: { x: number; y: number }): number =>
-    assertExists(dlcGuardQuadTree.find(p.x, p.y)).dlcGuard;
-
+  const { dlcGuard } = assertExists(
+    dlcGuardQuadTree.find(coordinates[0], coordinates[1]),
+  );
   const country = assertExists(context.countries.get(city.countryToken));
 
-  return {
-    type: 'Feature',
-    properties: {
-      type: 'city',
-      dlcGuard: getDlcGuard({ x: coordinates[0], y: coordinates[1] }),
-      stateName: country.name,
-      stateCode: country.code,
-      label: city.name,
-      tags: ['city'],
-    },
-    geometry,
-  };
+  return point(coordinates, {
+    type: 'city',
+    dlcGuard,
+    stateName: country.name,
+    stateCode: country.code,
+    label: city.name,
+    tags: ['city'],
+  });
 }
--- a/packages/clis/generator/commands/search.ts
+++ b/packages/clis/generator/commands/search.ts
@@ -361,7 +361,21 @@ function poiToSearchFeature(
     countriesById.get(closestNode.forwardCountryId) ??
     countriesById.get(closestNode.backwardCountryId);
   if (!country) {
-    logger.warn('unknown country id', closestNode.forwardCountryId, 'for', poi);
+    if (closestNode.forwardCountryId === closestNode.backwardCountryId) {
+      logger.warn(
+        'unknown country id',
+        closestNode.forwardCountryId,
+        'for',
+        poi,
+      );
+    } else {
+      logger.warn(
+        'unknown country ids',
+        [closestNode.forwardCountryId, closestNode.backwardCountryId],
+        'for',
+        poi,
+      );
+    }
     return [];
   }
 
--- a/packages/clis/generator/commands/search.ts
+++ b/packages/clis/generator/commands/search.ts
@@ -60,7 +60,7 @@ export const builder = (yargs: Argv) =>
     })
     .option('outputDir', {
       alias: 'o',
-      describe: 'Path to dir {usa,europe}-search.geojson should be written to',
+      describe: 'Path to dir {ats,ets2}-search.geojson should be written to',
       type: 'string',
       coerce: untildify,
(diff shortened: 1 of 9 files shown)


packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+    .option('extraLabels', {
+      alias: 'x',
+      describe: 'Path to extra-labels.geojson file (required for usa)',
+      type: 'string',
+      coerce: untildify,
+      demandOption: false,
+    })
+    .option('outputDir', {
+      alias: 'o',
+      describe: 'Path to dir {usa,europe}-search.geojson should be written to',
```
truckermudgeon:
```suggestion
      describe: 'Path to dir {ats,ets2}-search.geojson should be written to',
```

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+  logger.log('creating search.geojson...');
+
+  const tsMapData = normalizeDlcGuards(
+    readMapData(args.inputDir, args.map, {
+      mapDataKeys: searchMapDataKeys,
+    }),
+  );
+  if (args.map === 'europe') {
+    // HACK until europe is properly supported in dlc-guard normalization.
+    tsMapData.dlcGuardQuadTree = quadtree<{
```
truckermudgeon: Note to self: I should switch over all quadtrees to rtrees, just so there's only one type of spatial index being used.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+    };
+  }
+
+  extraLabels.features = extraLabels.features
+    .filter(f => {
+      return (
+        (f.properties.kind == null && f.properties.show == null) ||
+        ((f.properties.kind == 'town' || f.properties.kind == null) &&
+          f.properties.show === true &&
+          f.properties.text !== 'Golden Gate Bridge')
```
truckermudgeon: 🤔 it'd be nice if this conditional were simply `f.properties.kind === 'town' && f.properties.show`.

For the purposes of search, I only considered including scenery towns, not other labeled mileage targets like junctions, mountain passes, or bridges, because I want to be able to say that a particular POI is near a city/village (and not near a junction or pass).

I'll raise a PR in https://github.com/nautofon/ats-towns and see what nautofon thinks.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+        },
+        properties: {
+          text: v.name,
+          country: isoA2Ets2.get(v.state) ?? v.state,
+        },
+      })),
+    };
+  }
+
+  extraLabels.features = extraLabels.features
```
truckermudgeon: Rename this to `sceneryTowns`, and do this `usa`-specific filtering in the `if`/`else` above.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+          ? fromWgs84ToAtsCoords
+          : // ETS2 villages data is already in game coords.
+            (v: Position) => v;
+      f.geometry.coordinates = toGameCoords(
+        f.geometry.coordinates as [number, number],
+      );
+      return f;
+    });
+
+  const cityRTree = new RBush<
```
truckermudgeon, line 252: Move all this spatial-index building to a separate function, so `handler` can be focused on reading input and converting it into output.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+      state = assertExists(
+        tsMapData.countries.values().find(c => c.code === f.properties.country),
+        'unknown country code: ' + f.properties.country,
+      );
+    }
+    return {
+      ...f,
+      properties: {
+        dlcGuard: assertExists(
+          assertExists(context.dlcGuardQuadTree).find(
```
truckermudgeon: If we're always building a dlc guard quadtree, then this `assertExists` isn't necessary.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+      }),
+    ),
+  );
+  logger.success('done.');
+}
+
+function poiToSearchFeature(
+  poi: Poi,
+  context: MappedDataForKeys<typeof searchMapDataKeys> & {
+    dlcGuardQuadTree?: DlcGuardQuadTree;
```
truckermudgeon: This can be typed as always being present.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+
+  const closestNode = assertExists(nodeQuadTree.find(poi.x, poi.y)).node;
+  const countriesById = new Map<number, Country>(
+    context.countries.values().map(c => [c.id, c]),
+  );
+  const country =
+    countriesById.get(closestNode.forwardCountryId) ??
+    countriesById.get(closestNode.backwardCountryId);
+  if (!country) {
+    logger.warn('unknown country id', closestNode.forwardCountryId, 'for', poi);
```
truckermudgeon: Should also mention `backwardCountryId`?

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+  };
+
+  const containingCity = cityRTree
+    .search({
+      minX: poi.x,
+      minY: poi.y,
+      maxX: poi.x,
+      maxY: poi.y,
+    })
+    .at(0);
```
truckermudgeon, line 397: Either comment why we're using `at`, or declare an explicit type (and use `[0]` instead).

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+  const nearestCity = assertExists(cityQuadTree.find(poi.x, poi.y));
+  const cityProperties: { containingCity?: string; nearestCity?: string } = {
```
truckermudgeon, line 399: Currently, the "city" for a POI is one of these:
- `{ containingCity: string }`
- `{ nearestCity: string }`

And consuming code does this dance with `??`, _and_ has to check which field was used so that it knows if the city contains a POI or is merely near a POI 😕

Maybe it'd be better to have the "city" type be the following, instead:
- `{ city: string, relationship: 'contains' | 'near' }`

packages/clis/generator/commands/search.ts
truckermudgeon: Hmmm... maybe it _is_ ok to say that a particular POI is near a non-city/village:

image

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+    case 'road':
+      return [];
+    default:
+      throw new UnreachableError(poi);
+  }
+
+  if (city.stateCode !== baseProperties.stateCode) {
+    switch (properties.label) {
+      case 'El Capitan':
+        properties.nearestCity = 'Texas-Utah border';
```
truckermudgeon, line 484: 🤔 in Google Maps, some POIs don't have a city/town associated with them:

image

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+        properties.nearestCity = 'Teec Nos Pos';
+        break;
+      case 'Mesocco Castle':
+        properties.nearestCity = 'San Bernardino';
+        break;
+      case 'Chillon Castle':
+        properties.nearestCity = 'Vevey';
+        break;
+      case 'Karawanks Tunnel':
+        properties.nearestCity = 'Austria-Slovenia border';
```
truckermudgeon, line 499: Should city names even factor in? Are there a class of POIs where it's enough to say the state/country they're in, and that's it?

image

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+        break;
+      case 'New Europe Bridge':
+        properties.nearestCity = 'Romania-Bulgaria border';
```
truckermudgeon, line 502: 🤔 

image

Clicking on either POI marked as being in Romania or Bulgaria results in the same thing.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+      throw new UnreachableError(poi);
+  }
+
+  if (city.stateCode !== baseProperties.stateCode) {
+    switch (properties.label) {
+      case 'El Capitan':
+        properties.nearestCity = 'Texas-Utah border';
+        break;
+      case 'Monument Valley':
+        properties.nearestCity = 'Mexican Hat';
```
truckermudgeon, line 487: image

I'm not even sure what the best thing is to do, now 🫠

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+      properties,
+      geometry,
+    },
+  ];
+}
+
+function cityToSearchFeature(
+  city: City,
+  context: MappedDataForKeys<typeof searchMapDataKeys> & {
+    dlcGuardQuadTree?: DlcGuardQuadTree;
```
truckermudgeon: Same as above: make this required.

packages/clis/generator/commands/search.ts
```diff
@@ -0,0 +1,564 @@
...
+      stateName: country.name,
+      stateCode: country.code,
+      label: city.name,
+      tags: ['city'],
+    },
+    geometry,
+  };
+}
+
+function toDealerLabel(prefabPath: string): string {
```
truckermudgeon: This is copy/pasted from somewhere. Maybe it's better off exported from somewhere like `ui/map/labels`.

packages/clis/generator/geo-json/populated-places.ts
```diff
@@ -36,6 +36,11 @@ export const ets2IsoA2 = new Map([
   ['S', 'SE'],
 ]);
 
+/**
+ * Reverse map of {@link ets2IsoA2}.
```
truckermudgeon: Need to add a comment like `ets2IsoA2`s, saying that if a key isn't present, then it's assumed to already be an ISO A2 code.

truckermudgeon:
Good enough for now; will address:

- filtering extra-labels.geojson data
- changing how nearest- and containing- city data is represented
- dealing with POIs that are in one country/state, but closest to a city/scenery town in a different country/state

in the follow-up PR. The last two points might be better dealt with together, e.g., a POI's "city" data could be something like `{ closestCity: { name, stateCode, distance } }`, where 
- `distance: 0` denotes a containing city, and 
- consumers have all the info they need to decide what to show for a POI that's not close to any city, or is close to a city in a different state

packages/clis/generator/commands/search.ts
nautofon: I don't think I really understand the context here. FWIW, I'd say `'town'` is a reasonable default value for an unset `kind` attribute.

But the Golden Gate Bridge should, of course, have `kind: 'bridge'`. The generator shouldn't special-case that. I'll be sure to fix the metadata table at some point. If you want to do a PR, that might expedite things.

Or would you perhaps like write access to [ats-towns](https://github.com/nautofon/ats-towns) so that you could simply fix stuff like this yourself? I'd be fine with that. 🙂

truckermudgeon: > I'd say `'town'` is a reasonable default value for an unset `kind` attribute.

Ahhh... ok. That's good to know. Is it safe to say that the default value for `show` is `true`, too?

As for context: I'd like to be able to search for "scenery town", and get a list of scenery towns, only. I can definitely raise a PR for the Golden Gate Bridge (and any other things I encounter). Write-access would be fine, too (though I'd still like to run a PR by you, in case I fix other instances that might not be as straight forward).

nautofon: [Yes](https://github.com/nautofon/ats-towns/blob/80e9195402d8e3adb9ee8528488203a3b16e8710/label-metadata.md#show), the default of `show` is `true`.

Searching for category terms like that sounds interesting! 😀

I've now invited you as a collaborator to ats-towns. Feel free to push commits or raise PRs as you see fit. I also see the readme still needs an update. I hope to get around to that over the weekend, and will fix the Golden Gate Bridge entry at the same time if you haven't by then already.

PR #73, opened 2025-09-15
parser,generator: improve `search`'s city metadata

truckermudgeon:
This PR:
- updates `parser` so that company POIs include the company's `cityToken`
- updates `generator search` so that:
  - the city associated with a POI is guaranteed to be in the same state/country as the POI
  - the city metadata includes `distance`, as mentioned in [this comment](https://github.com/truckermudgeon/maps/pull/72#issuecomment-3289159525)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "packages/libs/*"
   ],
   "devDependencies": {
-    "@eslint/eslintrc": "^3.1.0",
-    "@eslint/js": "^9.12.0",
+    "@eslint/eslintrc": "^3.3.1",
+    "@eslint/js": "^9.35.0",
     "@typescript-eslint/eslint-plugin": "^8.8.0",
     "@typescript-eslint/parser": "^8.8.0",
     "@vitest/coverage-v8": "^2.1.8",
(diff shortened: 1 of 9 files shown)


packages/clis/parser/game-files/map-files-parser.ts
```diff
@@ -703,6 +703,7 @@ function postProcess(
           type: 'company',
           icon: item.token,
           label: companyName ?? item.token,
+          cityToken: item.cityToken,
```
truckermudgeon, line 706: There are now 3 pieces of information copied from the POI's associated `CompanyItem`. I'm starting to think I should just replace them with a `companyUid` field, and have POI consumers deal with linking so I don't have to re-parse everything whenever I decide to "forward" more fields 😕

truckermudgeon (commented):
I have learned two things from this PR:
- one should almost always specify the `maxResults` option when calling `PointRBush.findAll` (relying on the default `Infinity` value will end up returning all entries in the rtree)
- the `pois` array created by `parser` should probably rely more on indirection, and use foreign keys into the `parser`-created maps instead of copying values 🤔

package.json
```diff
@@ -10,8 +10,8 @@
     "packages/libs/*"
   ],
   "devDependencies": {
-    "@eslint/eslintrc": "^3.1.0",
-    "@eslint/js": "^9.12.0",
+    "@eslint/eslintrc": "^3.3.1",
+    "@eslint/js": "^9.35.0",
```
truckermudgeon, line 14: I thought maybe the unexplainable linting error in `point-rbush.ts` was a bug, and that upgrading would fix things. I was wrong.

PR #79, opened 2025-10-19
demo: add support for special events maps

truckermudgeon:
This PR updates the `demo` app with [`/brackenreach`](https://truckermudgeon.github.io/brackenreach) and [`/winterland`](https://truckermudgeon.github.io/winterland) pages that show the maps for the Halloween and Christmas special events.

https://github.com/user-attachments/assets/bb004201-7bb5-4c1e-a35a-7489015672f3

The maps can be accessed manually by visiting those URLs directly, or they can be accessed from the main UI when the special events are active and I remember to set the appropriate value in this line: https://github.com/truckermudgeon/maps/blob/bb0675d46cb8ed4734df8d51ba28108d11a36fa5/packages/apps/demo/src/index.tsx#L21

Note that the `.pmtiles` files for Brackenreach and Winterland were created as one-offs, because `parser` on the `main` branch doesn't know how to find their sector files. If/when updating the maps becomes painful enough, and/or SCS Software decides to add more portal-based maps to the game, I'll consider updating `parser` properly, so that special-events maps (and maybe even the Driving Academy map?) are always parsed when present.

### DIFF
--- a/packages/apps/demo/src/SpecialEventControl.tsx
+++ b/packages/apps/demo/src/SpecialEventControl.tsx
@@ -1,10 +1,23 @@
 import { Box, Link } from '@mui/joy';
 import { assertExists } from '@truckermudgeon/base/assert';
 import { christmasMapStyle, halloweenMapStyle } from '@truckermudgeon/ui';
+import { StyleSpecification } from 'maplibre-gl';
 import { useRef } from 'react';
 import { useControl } from 'react-map-gl/maplibre';
 
-export const eventMeta = {
+export type SpecialEvent = 'halloween' | 'christmas';
+
+export interface SpecialEventMeta {
+  emoji: string;
+  mapName: string;
+  url: `/${string}`;
+  centerLngLat: [number, number];
+  boundsDelta: number;
+  minZoom: number;
+  mapStyle: StyleSpecification;
+}
+
+export const eventMeta: Record<SpecialEvent, SpecialEventMeta> = {
   halloween: {
     emoji: '🎃',
     mapName: 'Brackenreach',
--- a/packages/apps/demo/src/SpecialEventControl.tsx
+++ b/packages/apps/demo/src/SpecialEventControl.tsx
@@ -1,9 +1,13 @@
 import { Box, Link } from '@mui/joy';
 import { assertExists } from '@truckermudgeon/base/assert';
 import { christmasMapStyle, halloweenMapStyle } from '@truckermudgeon/ui';
-import { StyleSpecification } from 'maplibre-gl';
+import type { StyleSpecification } from 'maplibre-gl';
 import { useRef } from 'react';
 import { useControl } from 'react-map-gl/maplibre';
+
+// TODO since `ui` components know about 'halloween' and 'christmas', then
+//  special events should be first-class concepts (and these event-related
+//  types should be moved to `lib/maps`, _and_ `parser` should handle them).
 
 export type SpecialEvent = 'halloween' | 'christmas';
 
--- a/packages/apps/demo/src/SpecialEventControl.tsx
+++ b/packages/apps/demo/src/SpecialEventControl.tsx
@@ -1,36 +1,46 @@
-import { IconButton } from '@mui/joy';
+import { Box, Link } from '@mui/joy';
 import { assertExists } from '@truckermudgeon/base/assert';
 import { useRef } from 'react';
-import { useControl, useMap } from 'react-map-gl/maplibre';
+import { useControl } from 'react-map-gl/maplibre';
+
+const eventMeta = {
+  halloween: {
+    emoji: '🎃',
+    mapName: 'Brackenreach',
+    url: '/brackenreach',
+  },
+};
 
 export const SpecialEventControl = (props: { specialEvent?: 'halloween' }) => {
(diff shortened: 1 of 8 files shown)


packages/apps/demo/src/SpecialEventMap.tsx
```diff
@@ -0,0 +1,91 @@
...
+  tileRootUrl: string;
+  specialEvent: 'halloween';
+}) => {
+  const { tileRootUrl, specialEvent } = props;
+  const { mode: _maybeMode, systemMode } = useColorScheme();
+  const mode =
+    _maybeMode === 'system' ? (systemMode ?? 'light') : (_maybeMode ?? 'light');
+
+  const map = ensureValidMapValue(localStorage.getItem('tm-map'));
+  const worldEmoji = map === 'usa' ? '🌎' : '🌍';
```
truckermudgeon, line 24: It's the little details that make me happy ☺️

packages/libs/ui/index.ts
```diff
@@ -16,31 +17,45 @@ export {
   StateCode,
 } from './SceneryTownSource';
 
+const baseMapStyle: StyleSpecification = {
+  version: 8,
+  // free font glyphs, required when adding text-fields.
+  // https://github.com/openmaptiles/fonts
+  glyphs: 'https://fonts.openmaptiles.org/{fontstack}/{range}.pbf',
+  // sources and layers are empty because they're declared as child
+  // components below.
```
truckermudgeon: A very outdated comment.

```suggestion
  // sources and layers are empty because they're declared as children
  // of consuming `MapGl` components.
```

packages/apps/demo/src/index.tsx
```diff
@@ -10,25 +10,56 @@ import {
...
+      <Route
+        path="street-view"
+        element={
+          <StreetViewDemo
+            tileRootUrl={tileRootUrl}
+            pixelRootUrl={pixelRootUrl}
+          />
+        }
+      />,
+    ].concat(
```
truckermudgeon, line 49: Just as map data for special events comes and goes, so do the special events map pages.

packages/apps/demo/src/SpecialEventMap.tsx
```diff
@@ -0,0 +1,91 @@
...
+        game={'ats'}
+        specialEvent={specialEvent}
+        mode={mode}
+        showSecrets={true}
+        enableIconAutoHide={false}
+      />
+      <NavigationControl visualizePitch={true} />
+      <FullscreenControl containerId={'fsElem'} />
+      <ModeControl />
+      <div style={{ position: 'absolute', top: 10, right: 52 }}>
```
truckermudgeon: I got lazy and didn't make this a MapLibre control.

packages/apps/demo/src/SpecialEventMap.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { ModeControl } from './ModeControl';
+
+export const SpecialEventMap = (props: {
+  tileRootUrl: string;
+  specialEvent: 'halloween';
+}) => {
+  const { tileRootUrl, specialEvent } = props;
+  const { mode: _maybeMode, systemMode } = useColorScheme();
+  const mode =
+    _maybeMode === 'system' ? (systemMode ?? 'light') : (_maybeMode ?? 'light');
```
truckermudgeon: 🤔 is there a way to avoid the double `?? 'light'`s?

packages/apps/demo/src/SpecialEventControl.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <div ref={ref} style={{ position: 'absolute', top: 0, right: 0 }}>
+      <div className={'maplibregl-ctrl maplibregl-ctrl-group'}>
+        <Link
+          component={'button'}
+          underline={'none'}
+          title={`Switch to ${meta.mapName} Map`}
+          justifyContent={'center'}
+          onClick={() => (window.location.href = meta.url)}
```
truckermudgeon, line 70: `component='button'` means we no longer can specify an `href` prop, and instead need to set an `onClick` handler 😔

packages/apps/demo/src/SpecialEventControl.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { specialEvent } = props;
+  if (specialEvent == null) {
+    return null;
+  }
+
+  const ref = useRef<HTMLDivElement>(null);
+  useControl(() => ({
+    onAdd: () => assertExists(ref.current),
+    onRemove: () => assertExists(ref.current).remove(),
+    getDefaultPosition: () => 'top-left',
```
truckermudgeon, line 57: Made this button a part of the `top-left` control group... which means that the top-left control (i.e., the `OmniBar`) needs to know about it 😕.

Figured this was the least-bad way to do things, compared to somehow making it part of the `NavigationControl`.

packages/apps/demo/src/index.tsx
truckermudgeon: Ok, I lied. For now, `/brackenreach` and `/winterland` will always be accessible, but the UI will only show links during the actual events (provided I remember to update + push changes to the `truckermudgeon.github.io` repo).

PR #78, opened 2025-10-14
demo: add support for Street Views

truckermudgeon:
This PR builds out the Photo Spheres feature added in https://github.com/truckermudgeon/maps/pull/65 and:
- adds support for [Street Views captured by `San_Sany4`](https://forum.scssoft.com/viewtopic.php?p=2064321#p2064321)
- replaces MapLibre GL JS's built-in URL hash functionality with a custom implementation that includes support for Street View URLs, e.g. https://truckermudgeon.github.io/#4/39/-108!cb3722a030c/24.31/-4.08/17.16
- promotes the Photo Spheres UI from an experimental feature to a normal, always-present-for-the-ATS-map feature

image

Note that the `generator` code to generate the panorama tiles and Street View GeoJSON for this feature still only exists on a local branch on my laptop, because the code is very messy and very much specific to my setup.

PR #80, opened 2025-10-22
demo: restore MapLibre-GL-JS-like hash history behavior

truckermudgeon:
The hash impl added in https://github.com/truckermudgeon/maps/pull/78 caused too many browser history states to be added, trapping users in a state where they couldn't back out of the page 😬

This commit fixes this by using `window.history.replaceState`, which is what [MapLibre uses](https://github.com/maplibre/maplibre-gl-js/blob/4aa5084c1223a6230d62214df1caadd74a8c3c3f/src/ui/hash.ts#L126) when updating URL hashes.

### DIFF
--- a/packages/apps/demo/src/OmniBar.tsx
+++ b/packages/apps/demo/src/OmniBar.tsx
@@ -110,17 +110,17 @@ export const OmniBar = (props: OmniBarProps) => {
         const zoom = e.target.getZoom();
         const bearing = Number(e.target.getBearing().toFixed(1));
         const pitch = Math.round(e.target.getPitch());
+        let hash;
         if (pitch) {
-          window.location.hash = `${zoom}/${lat.toFixed(3)}/${lng.toFixed(
+          hash = `#${zoom}/${lat.toFixed(3)}/${lng.toFixed(
             3,
           )}/${bearing}/${pitch}`;
         } else if (bearing) {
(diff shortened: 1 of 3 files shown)


packages/apps/demo/src/OmniBar.tsx
```diff
@@ -110,17 +110,17 @@ export const OmniBar = (props: OmniBarProps) => {
...
         } else if (bearing) {
-          window.location.hash = `${zoom}/${lat.toFixed(3)}/${lng.toFixed(
-            3,
-          )}/${bearing}`;
+          hash = `#${zoom}/${lat.toFixed(3)}/${lng.toFixed(3)}/${bearing}`;
         } else {
-          window.location.hash = `${zoom}/${lat.toFixed(3)}/${lng.toFixed(3)}`;
+          hash = `#${zoom}/${lat.toFixed(3)}/${lng.toFixed(3)}`;
         }
+        window.history.replaceState(window.history.state, '', hash);
```
truckermudgeon, line 123: Slight change in behavior, for the sake of consistency: users can no longer press back after using the search bar to jump to a point. But that didn't work quite as I'd expect: I can visit point A and B through the search bar, then manually pan to point C, then visit point D through the search bar. Pressing Back would take me to B (not C).

If I wanted full stateful histories (including, e.g., setting the contents of the search bar when pressing back/forward), then I'm guessing I'd have to switch to `pushState`.

PR #82, opened 2025-11-15
ETS2: Adjust water and countries borders for better fit with roads and footprints

Rares-Muntean:
I adjusted the `countries.geojson` and `water.geojson` from `resources` folder so they align better with the roads and footprints generated by the CLI tool.
These layers were manually adjusted in QGIS to create a cleaner version of the map with the road network. It's not perfect, but the borders and water areas now blend more naturally. Differences might not be immediately noticeable, but they improve the overall alignment.

I wasn’t fully certain whether these refinements were expected in the project, but I committed them because they improve the visual integration.

Notes:

- The refinements focus on the European region only.


truckermudgeon:
I was just thinking about the need to update these files 😅 

It's been on my TODO list for a very long time, and over the past week I've even started experimenting with semi-automated ways of updating this data so that it's closer to what's seen in the in-game map.

I'll take a closer look at your changes in the next few days or so. In the meantime, if there's any before + after screenshots of QGIS that you can add to the PR description, or any more details about your manual process (e.g., did you try adjusting borders by tracing over a screenshot of the in-game map?), then that would greatly be appreciated.

Also: you'll need to make sure `prettier` is run on the `geojson` files before committing, otherwise the linter will complain (that's usually enforced by the Husky commit hooks, but I guess they've been disabled on your machine).

Rares-Muntean:
> Also: you'll need to make sure `prettier` is run on the `geojson` files before committing, otherwise the linter will complain (that's usually enforced by the Husky commit hooks, but I guess they've been disabled on your machine).

I'll look into it, thanks!

> or any more details about your manual process

Well as I mentioned, it’s a very raw manual process. I didn’t do anything fancy, just dragged vertices around until it looked okay visually.

So first thing that I did was to use the **Affine Transform** tool on the `water.geojson` with these options:

- `Translation (x-axis): 0.100000`
- `Translation (y-axis): -0.040000`

And after that I edited the areas where the footprints and roads seemed to go over streets using the **vertex tool**. I just dragged individual vertex points to make the water polygons smaller (dragged them in). I Probably could've selected multiple points at once to move and drag them nicely without breaking the original shape but that didn't go through my mind.


> did you try adjusting borders by tracing over a screenshot of the in-game map

Not really, even some areas of the in-game map are inaccurate. I used the `ets2.pmtiles` layer as a reference in **QGIS** instead.

Here is a quick before and after of the map layers.

Close-up:
closeup-example

And the whole map:
overall-example
(the last layer is the one modified and the one above is the original)

As you can see the differences are subtle but i think it looks good with mapLibre:

{1E299A78-86D6-4D88-BC81-E228545EE228}

truckermudgeon:
Those before/afters do look good, thanks for posting them!

> So first thing that I did was to use the Affine Transform tool on the water.geojson with these options

Looks like using the transformation tool has affected the country + water borders for the entire world, not just the ETS2 areas 😬 

I don't think the rest of the world map should change if we're only interested in improving the ETS2 map... is there a way to revert those changes, such that the changes in this PR only affect the parts of the world covered by the ETS2 map?

Rares-Muntean:
I will look into it! Thanks for bringing that up

Rares-Muntean:
If you can take a look to see if it's alright now? I know it's a little bit weird with the map split in half (tho I don't think it should be visible) But it wouldn't let me modify parts of the water otherwise.

Also i've commited with syncpack and husky tool so it should be alright I think

truckermudgeon:
Hey @MunteanRares! Sorry for the delay in review: just got back from a trip, and I've got some time off due to a holiday here in the US. I should be able to take a look this weekend (and see how the new borders look with the Nordic Horizons release, too).

packages/clis/generator/resources/countries.geojson
```diff
@@ -1,14 +1,18 @@
 {
   "type": "FeatureCollection",
-  "name": "countries",
+  "name": "ets2-countries",
```
truckermudgeon: Nit: this should probably still say named as "countries", because the features in the collection include country borders in both ATS and ETS2.

packages/clis/generator/resources/water.geojson
truckermudgeon, line 1: Any way to get rid of the seam running down the middle of the world?

Image

truckermudgeon (commented):
Had a couple of comments/questions. The biggest blocker is the seam that came from splitting the world.

Also: do you have any plans / interest in adjusting the GeoJSON to account for the latest changes in Nordic Horizons?

Image

packages/clis/generator/resources/water.geojson
nautofon: The entire eastern hemisphere has been moved by about 0.1° towards the ESE. Assuming that this intended, the easiest solution seems to be to manually adjust the vertices along the seam so that it appears closed, then run the Dissolve tool (in QGIS → Vector → Geoprocessing) with the "Keep disjoint features separate" option enabled.

Afterwards, any unneeded vertices along the now-closed seam can be removed manually if desired. Furthermore, the eastern edge of the data should be fixed so that it doesn't cross the 180th meridian, and that the eastern and western data edges match up.

I had a go at this and will open a PR on https://github.com/MunteanRares/maps with the result.

truckermudgeon:
An update from my side of things:

First, thank you @MunteanRares for taking the time to _improve the world_ 😉 (and thank you, @nautofon, for helping close the seam, _and_ teaching me about the Dissolve tool through your comments).

I started taking a closer look at the changes, and some of them I like (e.g., the changes near Poland and Lithuania), and some of them I'm a little hesitant to make, e.g., this change that makes some Italian ferry lines run through land:

image

(In the case of the ferry line issue, I think that's mainly caused by the shifting of everything 0.1° towards the ESE.)

My current plan is to:

1. perform my own updates to `water.geojson`, using a vectorized form of the map background that ships with ETS2 (vectorization done with [this tool](https://www.visioncortex.org/vtracer/)). This will address the horribly outdated borders in the NH DLC:
image

2. then, apply some of the changes you made in this PR to the updated `water.geojson`. Fortunately, I can see all the tweaks you made by overlaying some layers in QGIS:
  image <br>Applying them selectively shouldn't be too hard: I'm pretty sure it'll just involve playing with `Vector > Geoprocessing Tools`.

3. lastly, performing steps 1 and 2 again, but to `countries.geojson`

----

Looks like I can update this PR directly:

image

so expect to see updates (and a merged PR) Soonish™ 🙂

packages/clis/generator/resources/water.geojson
truckermudgeon: Resolved a different way (using boolean operations to modify just the ETS2 area of the map).

truckermudgeon:
Sooooo... that took me longer than I wanted it to 🫠 

Many of the coastline changes @MunteanRares suggested were no longer needed when I updated the ETS2 data to better match the in-game map,e.g., the coast of Portugal:

image

(this is a map of the updated ETS2 data, with the pink areas being @MunteanRares's tweaks to the old ETS2 data)

But there were still a handful or two of @MunteanRares's changes that were still preserved (e.g., the changes to Corsica)

I hope to never have to do this again for a while, because it's so time consuming: "redoing" all of the ETS2 data meant I had to re-digitize many of the country borders, and because I'm still a noob with QGIS and I ended up making a few mistakes that led me to lose a lot of work (pro-tip: MAKE THOSE TEMPORARY LAYERS PERMANENT ASAP 🫠 🫠)...

...but I know I'll end up having to update the ATS side of things for Louisiana (but that should hopefully go a lot smoother 🤞).

nautofon:
Nice work, both of you. 🙂

PR #88, opened 2026-01-08
parser: parse sectors using `binary-parser` library instead of `restructure` library

truckermudgeon:
This PR updates `parser`'s sector parsing to use [`binary-parser`](https://github.com/keichi/binary-parser).

`binary-parser`'s APIs differ from `restructure`'s:

```ts
// binary-parser
const fooParser = new Parser()
  .uint16le('field1')
  .uint16le('field2')
  .uint8('arrayFieldCount')
  .array('arrayField', {
    type: 'uint32le',
    length: 'arrayFieldCount',
  });

// restructure
const fooParser = new r.Struct({
  field1: r.uint16le,
  field2: r.uint16le,
  arrayField: new r.Array(r.uint32le, r.uint8),
});
```

I prefer `restructure`'s builder APIs over `binary-parser`'s, so I've gone with the approach of writing adapters so that `sector-parser.ts`'s builder code stays the same.

Switching over to `binary-parser` noticeably speeds things up: parsing ATS sectors is almost 3x faster, cutting the entire parsing process by about half.

```
// old
✔ parsed 1722 sector files in 172.325 seconds
npx parser -i  -o out/parser --dryRun  256.92s user 34.84s system 134% cpu 3:36.870 total

// new
✔ parsed 1722 sector files in 67.252 seconds
npx parser -i  -o out/parser --dryRun  137.52s user 10.48s system 151% cpu 1:37.400 total
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6404,6 +6404,15 @@
       "integrity": "sha512-aOk0SxEon5LF9cMxQFViSKb4qccG6rs7XKyMXIb1J8f8LA2acTIWnHdT0IOTe4gYBbqgjdbuTZ5f+UP+vlh4Mw==",
       "license": "MIT"
     },
+    "node_modules/binary-parser": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/binary-parser/-/binary-parser-2.3.0.tgz",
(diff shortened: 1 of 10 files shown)


truckermudgeon (commented):
Note: `dds-parser`, `prefab-ppd-parser` and `scs-archive` have _not_ been migrated to use `binary-parser` because:
- it requires me to do more adapting than I want to (e.g., of `restructure`'s `Pointer` API, and adding serialization/encoding support)
- I suspect the gains from parsing those formats are tiny

PR #89, opened 2026-01-13
demo: add 'show as dashed lines' display option for secret roads

truckermudgeon:
https://github.com/user-attachments/assets/3eb79fde-53c2-49f9-b8f1-9453c990d90c

### DIFF
--- a/packages/libs/ui/GameMapStyle.tsx
+++ b/packages/libs/ui/GameMapStyle.tsx
@@ -276,7 +276,14 @@ export const GameMapStyle = (props: GameMapStyleProps) => {
         ]}
         layout={roadLineLayout}
         paint={{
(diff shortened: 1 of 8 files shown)


packages/libs/ui/GameMapStyle.tsx
```diff
@@ -257,6 +261,46 @@ export const GameMapStyle = (props: GameMapStyleProps) => {
...
+            10,
+            1,
+            14,
+            2,
+            16,
+            3,
+          ],
+          ...(showSecrets === 'showAsDashed'
+            ? { 'line-dasharray': [2, 2] }
+            : {}),
```
truckermudgeon, line 301: I thought I could just specify a value of `undefined`, thinking that was the same thing as _not_ specifying the `line-dasharray` property... but doing that led to a bunch of console errors 😬

Image

PR #90, opened 2026-01-30
generator: use connected roads to improve prefab-road generation

truckermudgeon:
(I've got _a bunch of really old code_ that I haven't published yet. I'm trying to change that by just pushing code out to GitHub, with minimal review. This PR is based on code I wrote in August of 2024 😬)

This PR tries to improve the way LineStrings for road-like prefabs are generated by taking into account whether or not connecting roads are divided (i.e., if connecting roads have an `offset`).

On balance, I think the results are better: there are less "chain-link" like roads that appear in the map, at the cost of stranger-looking three/four-way intersections and slightly-disconnected roads.

| Before  | After |
| ------------- | ------------- |
| image  | image |

I plan on re-visiting the three/four-way intersections In The Future™, modeling them in a way that matches [MapBox's guidance](https://web.archive.org/web/20230602212339/https://labs.mapbox.com/mapping/mapping-for-navigation/modeling-intersections-for-map-navigation/), e.g.: 

image

### DIFF
--- a/packages/clis/generator/geo-json/map.ts
+++ b/packages/clis/generator/geo-json/map.ts
@@ -1249,6 +1249,7 @@ function augmentWithRoadContext(
   prefab: Prefab,
   prefabDescription: PrefabDescription,
   nodes: ReadonlyMap<bigint, Node>,
+  // TODO replace these with a Map<bigint, RoadLook>, keyed by prefab node uids.
   roadLookMap: ReadonlyMap<string, RoadLook>,
   roadQuadTree: Quadtree<{ x: number; y: number; roadLookToken: string }>,
 ): {
@@ -1258,6 +1259,8 @@ function augmentWithRoadContext(
   const tx = (pos: Position) =>
     toMapPosition(pos, prefab, prefabDescription, nodes);
 
+  // HACK clone `prefabDescription`, so we can mutate its clone by augmenting
+  // its RoadMapPoints with contextual offset info.
   const augmentedPd = JSON.parse(
     JSON.stringify(prefabDescription),
   ) as PrefabDescription;
--- a/packages/clis/generator/geo-json/map.ts
+++ b/packages/clis/generator/geo-json/map.ts
@@ -17,6 +17,7 @@ import {
   toMapPosition,
   toRoadStringsAndPolygons,
 } from '@truckermudgeon/map/prefabs';
+import { getRoadType } from '@truckermudgeon/map/roads';
 import type {
   AtsMapGeoJsonFeature,
   City,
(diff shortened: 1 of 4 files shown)


packages/clis/generator/geo-json/map.ts
```diff
@@ -1220,25 +1245,81 @@ function countryToFeature(country: Country): CountryFeature {
   };
 }
 
+function augmentWithRoadContext(
+  prefab: Prefab,
+  prefabDescription: PrefabDescription,
+  nodes: ReadonlyMap<bigint, Node>,
+  roadLookMap: ReadonlyMap<string, RoadLook>,
+  roadQuadTree: Quadtree<{ x: number; y: number; roadLookToken: string }>,
```
truckermudgeon, line 1254: TBH, I can't remember why I decided to use a Quadtree instead of just passing in the `roads` and `roadLooks` maps... maybe it was slightly easier than looking up a connecting `road` by prefab node UID, and then looking up the `roadLook` for that road? 

If that's the case, maybe it'd be better to build a map of prefab node UIDs to roadLooks, rather than relying on a spatial index 😕

packages/clis/generator/geo-json/map.ts
```diff
@@ -1220,25 +1245,81 @@ function countryToFeature(country: Country): CountryFeature {
...
+): {
+  roadStrings: RoadString[];
+  polygons: Polygon[];
+} {
+  const tx = (pos: Position) =>
+    toMapPosition(pos, prefab, prefabDescription, nodes);
+
+  const augmentedPd = JSON.parse(
+    JSON.stringify(prefabDescription),
+  ) as PrefabDescription;
```
truckermudgeon, line 1266: Deserves a comment about being a hacky clone.

packages/clis/generator/geo-json/map.ts
```diff
@@ -1220,25 +1245,81 @@ function countryToFeature(country: Country): CountryFeature {
...
+    //console.log('tweaking', prefab.uid, matchingRoadPoint, roadLook);
+    matchingRoadPoint.offset = roadLook.offset ?? 0;
+    if (matchingRoadPoint.offset > 0) {
+      const maxLanes = Math.max(
+        roadLook.lanesLeft.length,
+        roadLook.lanesRight.length,
+      );
+      matchingRoadPoint.lanesLeft = maxLanes;
+      matchingRoadPoint.lanesRight = maxLanes;
+    }
```
truckermudgeon, line 1299: Huh. I wonder why the road point's left/right lane counts are both being set to the max lane count of the connected road 🤔

packages/libs/map/roads.ts
```diff
@@ -0,0 +1,50 @@
...
+    roadType = 'local';
+  } else if (lanes.some(l => l.includes('tram'))) {
+    roadType = 'tram';
+  } else if (lanes.some(l => l.includes('train'))) {
+    roadType = 'train';
+  }
+  return roadType;
+}
+
+export function getLaneSpeedClass(look: RoadLook): LaneSpeedClass {
```
truckermudgeon, line 36: Not used in this PR, but will be used in updates to graph-generation and route-finding code.

Issue #87, opened 2025-12-27
Bugs


truckermudgeon:
Another issue serving as a TODO list, similar to #86.

- [ ] Ferry routes don't respect the selected DLC options
  - partially fixed by #119 
- [ ] Ferry route durations aren't calculated using `FerryConnection::time` (see [this comment](https://github.com/truckermudgeon/maps/pull/91/changes#r2748351887))
- [x] Calculation of UK coords needs to be redone (see [this](https://forum.scssoft.com/viewtopic.php?p=2061509#p2061509) report) ([done](https://github.com/truckermudgeon/maps/pull/140))
- [x] ~Companies east of St. Louis, MO, show up on the map~ ([IL released](https://store.steampowered.com/app/3486960/American_Truck_Simulator__Illinois/); no longer relevant)
- [x] Strafford, MO is signed, but doesn't appear on the map ([done](https://github.com/nautofon/ats-towns/pull/11))
- [ ] Gas Stations unusable by trucks show up on the map
- [ ] Border crossing icon is broken (see [this comment](https://forum.scssoft.com/viewtopic.php?p=2140538#p2140538))

PR #91, opened 2026-01-30
generator,map: add support for finding 'fastest' route

truckermudgeon:
This PR:
- updates the `generator graph` command to include duration information in the routing graph: every `Neighbor` entry now has:
```ts
  /** The estimated duration to travel to this neighbor's node (seconds). */
  readonly duration: number;
```
- updates `findRoute` with the ability to find a route based on the shortest duration (i.e., the `fastest` route)
  - successfully returned `Route`s also include the duration of the entire route, so that it can be used in UI that displays information about the routes available from point A to point B:

image

### DIFF
--- a/packages/clis/generator/graph/graph.ts
+++ b/packages/clis/generator/graph/graph.ts
@@ -18,9 +18,11 @@ import {
   fromAtsCoordsToWgs84,
   fromEts2CoordsToWgs84,
 } from '@truckermudgeon/map/projections';
+import { getLaneSpeedClass } from '@truckermudgeon/map/roads';
 import type { Direction } from '@truckermudgeon/map/routing';
 import type {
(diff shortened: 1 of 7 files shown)


packages/clis/generator/graph/graph.ts
```diff
@@ -1233,15 +1299,18 @@ function createDebugLineString(
   );
 }
 
+// assumes speed of 30 mph
 function createNeighbor(
```
truckermudgeon, line 1303: Oof... this is used when creating edges between company / facility nodes and roads / prefabs... which is probably ok.

But it's also used when creating edges between ferries, where a speed of 30 mph might not be a good assumption.

Added an entry to https://github.com/truckermudgeon/maps/issues/87.

PR #107, opened 2026-03-13
navigation: fix some connection and logging issues

truckermudgeon:
Got a few reports in the [Navigator thread](https://forum.scssoft.com/viewtopic.php?p=2118044) about unexpected disconnections, e.g., immediately after entering a pairing code.

I'm not sure if this PR will fix those specific cases, but I do know it'll fix at least a couple:

- if, while connected to Navigator, you reload your browser 10 times, the 10th reload will result in a perpetual loading screen because event listeners installed from the previous reloads weren't being uninstalled
- the "concurrent connections by ip address" limiting is broken: resetting / cooldown wasn't working properly

Also, turns out the `loggingMiddleware` doesn't run when a `subscription` errors... such errors need to be caught inline (`loggingMiddleware` only catches errors during setup / initialization of the subscription 🤦). This PR fixes that, too.

### DIFF
--- a/packages/apis/navigation/trpc/routers/navigator-router.ts
+++ b/packages/apis/navigation/trpc/routers/navigator-router.ts
@@ -406,6 +406,7 @@ export const navigatorRouter = router({
   subscribeToDevice: navigatorSessionProcedure
     .use(subscriptionLimitMiddleware(1))
     .subscription(async function* ({
+      path,
       ctx,
       signal,
     }): AsyncGenerator<ActorEvent, void, void> {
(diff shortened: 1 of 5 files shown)


packages/apis/navigation/trpc/routers/navigator-router.ts
```diff
@@ -420,18 +421,36 @@ export const navigatorRouter = router({
...
+          },
+          request: {
+            type: ctx.type,
+            clientId: ctx.clientId,
+          },
+          error:
+            err instanceof Error ? `${err.name}: ${err.message}` : String(err),
+        });
+      } finally {
+        unsubscribe();
```
truckermudgeon, line 453: TODO: update `apps/navigator` to detect subscriptions that have ended, and to display appropriate UI (e.g., "lost connection. try reloading").

PR #113, opened 2026-03-21
navigator: implement north lock

truckermudgeon:
This PR:
- changes the compass widget from being text-only to being a graphical one, that looks similar to Apple Maps'
- adds tap handlers to the compass, to toggle between north-lock and north-unlock modes
  - note: north-unlock syncs the compass to the bearing of the _map_, not of the truck

https://github.com/user-attachments/assets/a1770a87-87d8-4242-ae5f-72703a91a2f2

### DIFF
--- a/packages/apps/navigator/src/create-app.tsx
+++ b/packages/apps/navigator/src/create-app.tsx
@@ -290,7 +290,7 @@ export function createApp({
         store.mapLoaded = true;
         controller.onMapLoad(map, marker);
         controller.startListening(store, appClient);
-        controlsController.startListening(controlsStore, appClient);
+        controlsController.startListening(controlsStore, appClient, map);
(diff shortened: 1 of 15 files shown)


packages/apps/navigator/src/create-app.tsx
```diff
@@ -264,6 +266,19 @@ export function createApp({
   });
   const _Controls = () => (
     <Controls
+      onCompassClick={action(() => {
```
truckermudgeon, line 269: 🤔 it's possible to tap on this while searching for POIs... and searching for POIs always clears the map's bearing + pitch... so the effects of tapping take place, but it's not clear what those effects are until the search UI is complete.

Consider ignoring taps when the sheet nav is visible?

PR #116, opened 2026-04-04
navigator: make route lines look nicer

truckermudgeon:
This PR does a handful of things to make route lines look nicer. It:
- colors the traveled part of a route gray (the route now reflects progress)
- renders a circle at the start and end of routes
- animates the route previews shown during search flows
- makes route lines thinner at lower zoom levels

https://github.com/user-attachments/assets/5c484078-1941-46b2-bd30-7fd682c87ee9

<br>
<br>
<br>

The whole "progress" part was interesting: my initial attempt used a single line and a line-gradient, where line-gradient progress values are normalized from `[0, 1]`, representing some point along the entire route.

For very long routes, a normalized progress value isn't accurate enough to mark the true progress along the route... rounding errors lead to progress appearing faster than the truck:

https://github.com/user-attachments/assets/656945cc-9a78-4796-9c29-bdb2a5b19cfa

My workaround was to render two lines:
- the line that represents the entire route, which may have inaccurate progress marked
- another, much shorter line, that represents the current routing "step"
  - this second line still uses a line-gradient with a normalized progress value from `[0, 1]`, but it's far less susceptible to rounding errors
  - this second line is rendered on top of the "main" route line

### DIFF
--- a/packages/apps/navigator/src/components/RoutesStyle.tsx
+++ b/packages/apps/navigator/src/components/RoutesStyle.tsx
@@ -43,36 +43,63 @@ export const RoutesStyle = () => {
             }),
           }}
         />
+      </Source>
+      <Source
+        id={'activeRouteStep'}
+        type={'geojson'}
+        lineMetrics={true}
+        data={
+          {
+            type: 'FeatureCollection',
+            features: [],
+          } as GeoJSON.FeatureCollection
+        }
+      >
         <Layer
-          id={'activeRouteIconsLayer'}
-          type={'symbol'}
-          layout={{
-            'icon-image': '{sprite}',
-            'icon-allow-overlap': true,
-          }}
-          minzoom={10}
-          filter={[
-            'all',
-            ['==', ['geometry-type'], 'Point'],
-            ['==', ['get', 'type'], 'traffic'],
-          ]}
-        />
-        <Layer
-          id={'activeRouteStartLayer'}
-          type={'circle'}
+          id={'activeRouteStepLayer'}
+          source={'activeRouteStep'}
+          type={'line'}
           paint={{
-            'circle-radius': startPointWidth,
-            'circle-color': '#fff',
-            'circle-stroke-width': 2.5,
-            'circle-stroke-color': '#888',
+            'line-width': routeLineWidth,
+            'line-opacity': 1,
+            'line-gradient': lineGradientExpression({
+              lineType: 'line',
+              progress: 0,
+            }),
           }}
-          filter={[
-            'all',
-            ['==', ['geometry-type'], 'Point'],
-            ['==', ['get', 'type'], 'start'],
-          ]}
         />
       </Source>
+      <Layer
+        id={'activeRouteIconsLayer'}
+        source={'activeRouteStep'}
+        type={'symbol'}
+        layout={{
+          'icon-image': '{sprite}',
+          'icon-allow-overlap': true,
+        }}
+        minzoom={10}
+        filter={[
+          'all',
+          ['==', ['geometry-type'], 'Point'],
+          ['==', ['get', 'type'], 'traffic'],
+        ]}
+      />
+      <Layer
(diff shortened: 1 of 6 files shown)


packages/apps/navigator/src/controllers/app.ts
```diff
@@ -533,7 +567,7 @@ export class AppControllerImpl implements AppController {
       }
 
       const { speed, position, heading } = gameState;
-      const { position: center, bearing } = toPosAndBearing({
+      const { position: _center, bearing } = toPosAndBearing({
```
truckermudgeon: revert

packages/apps/navigator/src/controllers/app.ts
```diff
@@ -543,6 +577,7 @@ export class AppControllerImpl implements AppController {
           heading,
         },
       });
+      const center = _center;
```
truckermudgeon: ditto

packages/apps/navigator/src/components/RoutesStyle.tsx
```diff
@@ -0,0 +1,269 @@
...
+            'line-opacity': 1,
+            'line-gradient': lineGradientExpression({
+              lineType: 'case',
+              progress: 0,
+            }),
+          }}
+        />
+        <Layer
+          id={'activeRouteStepLayer'}
+          source={'activeRouteStep'}
```
truckermudgeon: unnecessary

packages/apps/navigator/src/components/RoutesStyle.tsx
```diff
@@ -0,0 +1,269 @@
...
+            'line-gradient': lineGradientExpression({
+              lineType: 'line',
+              progress: 0,
+            }),
+          }}
+        />
+      </Source>
+      <Layer
+        id={'activeRouteIconsLayer'}
+        source={'activeRouteStep'}
```
truckermudgeon: TIL: `<Layer>` components don't need to be nested in a `<Source>`.

packages/apps/navigator/src/components/SlippyMap.tsx
```diff
@@ -30,6 +34,51 @@ const extents = {
   ].flat() as Extent,
 };
 
+const lineColors = {
```
truckermudgeon: `RoutesStyle.tsx` is a less-worse place for these to live.

packages/apps/navigator/src/controllers/app.ts
```diff
@@ -279,6 +310,9 @@ export class AppControllerImpl implements AppController {
     bottom: 0,
   };
   private offset: [number, number] = [0, 0];
+  private lastRenderedActiveStepLine:
```
truckermudgeon, line 313: It's possible that MapLibre's GeoJSON is fast enough that this field isn't needed... but I'm too lazy to benchmark and check.

packages/apps/navigator/src/controllers/app.ts
```diff
@@ -691,24 +722,158 @@ export class AppControllerImpl implements AppController {
...
+      );
+
+    this.lastRenderedActiveStepLine = store.activeStepLine.line;
+
+    const progress = clamp(
+      distanceTraveled / store.geoJsonRoute.featureLength,
+      0,
+      1,
+    );
+    this.map
```
truckermudgeon, line 858: 🤔 does the main line need to be updated as frequently as the current-step line?

packages/apps/navigator/src/components/RoutesStyle.tsx
truckermudgeon: Also: this is the wrong `source` reference 😬

packages/apps/navigator/src/controllers/app.ts
truckermudgeon: Maybe it only needs to change every time the active-step changes.

An optimization for A Later Time™

PR #119, opened 2026-04-08
parser,generator,demo: add support for ETS2 `dlcGuard`s

truckermudgeon:
This PR changes how `dlcGuard: 0` values for certain map items are calculated:
- **Before**: an ATS-only calculation, that relied on the fact that each ATS DLC is associated with one other State
- **After**: a calculation that works for both ATS + ETS2: use the `.scs` file that a map item appears in to tell what DLC it's associated with

This means that I can finally implement a DLC picker for ETS2, similar to the DLC picker for ATS:

image

### DIFF
--- a/packages/apps/demo/src/Legend.tsx
+++ b/packages/apps/demo/src/Legend.tsx
@@ -115,9 +115,8 @@ export interface LegendProps {
     enableAutoHiding: boolean;
     onAutoHidingToggle: (newValue: boolean) => void;
   };
+  dlcs: ListProps<number> & { map: 'usa' | 'europe' };
   advanced: AdvancedOptionsProps;
-  atsDlcs: ListProps<AtsSelectableDlc>;
-  ets2Dlcs: ListProps<Ets2SelectableDlc>;
 }
 export const Legend = (props: LegendProps) => {
   const ref = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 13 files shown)


packages/apps/demo/src/Legend.tsx
```diff
@@ -108,8 +115,8 @@ export interface LegendProps {
     enableAutoHiding: boolean;
     onAutoHidingToggle: (newValue: boolean) => void;
   };
+  dlcs: ListProps<number> & { map: 'usa' | 'europe' };
```
truckermudgeon, line 118: Something to consider:

`map` is only used to determine which set of checkbox labels to use. That information could (should?) be part of `ListProps::allItems`.

packages/clis/generator/dlc-guards.ts
```diff
@@ -48,34 +48,41 @@ export type DlcGuardMappedData = MappedDataForKeys<typeof dlcGuardMapDataKeys>;
...
     .x(e => e.x)
     .y(e => e.y);
   const unknownDlcGuards = new Set<number>();
 
-  // returns a normalized dlc guard, or undefined if dlcGuard cannot / should
-  // not be normalized.
+  // returns a normalized dlc guard (i.e., a guard value where `0` means "base
+  // map content") for the given `nodeUids`, or `undefined` if dlcGuard
+  // cannot / should not be normalized.
+  // TODO uhhhhh... wat. this needs to be re-examined.
```
truckermudgeon, line 79: Yeah... this code is very old, and I don't remember what I was thinking when I wrote it 🫠

packages/clis/parser/game-files/combined-entries.ts
```diff
@@ -56,11 +56,14 @@ class CompositeDirectory implements DirectoryEntry {
   readonly hash: bigint;
   readonly files: readonly string[];
   readonly subdirectories: readonly string[];
+  readonly scsSource = 'composite';
+  readonly scsSources: readonly string[];
```
truckermudgeon, line 60: It may be useful to know what `.scs` files a map sector entry belongs to, if/when I get around to adding mod support.

PR #128, opened 2026-04-30
generator: make use of turf helpers to build GeoJSON features

truckermudgeon:
I can't remember why I think initially avoided using the turf helpers... but I _think_ it's because I didn't want to introduce a dependency on `@turf/helpers` just so I could avoid writing object literals.

Time has passed, and I've reconsidered.

This PR updates GeoJSON object literals with turf-built equivalents.

### DIFF
--- a/packages/clis/generator/geo-json/map.ts
+++ b/packages/clis/generator/geo-json/map.ts
@@ -54,6 +54,7 @@ import type {
   WithToken,
 } from '@truckermudgeon/map/types';
 import * as turf from '@turf/helpers';
+import { lineString, point, polygon } from '@turf/helpers';
 import lineOffset from '@turf/line-offset';
 import type { Quadtree } from 'd3-quadtree';
 import { quadtree } from 'd3-quadtree';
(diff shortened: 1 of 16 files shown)


packages/clis/generator/geo-json/map.ts
```diff
@@ -1047,19 +1037,14 @@ function areaToFeature(
...
       secret: !!area.secret,
       zIndex: area.drawOver ? 1 : 0,
       color: area.color,
-    },
-    geometry: {
-      type: 'Polygon',
-      coordinates: [points],
-    },
+    }),
+    id: area.uid.toString(16),
```
truckermudgeon, line 1047: I'd like to write:

```ts
return polygon(
  [points], 
  { /* ...properties */ }, 
  { id: area.uid.toString(16),
);
```

but I get type errors because the expected return type of `areaToFeature` (and a few other `xxxToFeature` functions) looks something like:
```ts
{
  type: 'Feature',
  // ...geometry
  // ...properties,
  id: string
}
```

But `polygon` (and the other turf feature-builders) has a return type that looks something like:

```ts
{
  type: 'Feature',
  // ...geometry
  // ...properties,
  id?: string | number | undefined;
}
```

Where the type of `id` is too wide... so that's why there's this weird destructure-then-manually-specify-id dance.

PR #129, opened 2026-04-30
navigation,navigator: split composite roundabout maneuvers into entry/exit steps

truckermudgeon:
Currently, entering + exiting a composite roundabout was treated as a single maneuver, e.g.: "enter the roundabout and take the 2nd exit".

This PR makes it so that composite roundabouts have separate entry + exit steps, e.g.:
1. Enter the roundabout and take the 2nd exit
2. Exit the roundabout

https://github.com/user-attachments/assets/c34ffe24-dd5c-4731-92de-554a4ff4bcd3

Which seems to match what Google Maps does for larger roundabouts, e.g.:

https://github.com/user-attachments/assets/2ff350c6-808b-4456-9a9d-f437668b8598

### DIFF
--- a/packages/apis/navigation/domain/actor/route-step-builder.ts
+++ b/packages/apis/navigation/domain/actor/route-step-builder.ts
@@ -12,6 +12,7 @@ import {
 import { UnreachableError } from '@truckermudgeon/base/precon';
 import type { MappedDataForKeys } from '@truckermudgeon/io';
 import { ItemType } from '@truckermudgeon/map/constants';
+import { getCommonItem } from '@truckermudgeon/map/get-common-item';
 import { getLineString } from '@truckermudgeon/map/linestring';
 import {
   calculateLaneInfo,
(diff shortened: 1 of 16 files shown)


packages/apis/navigation/domain/actor/route-step-builder.ts
```diff
@@ -49,59 +56,97 @@ type RouteStepMappedData = MappedDataForKeys<
...
+    const item = getCommonItem(
+      startNode.uid,
+      endNode.uid,
+      this.tsMapData,
+      this.lookup,
+    );
+
+    // If we're inside a multi-prefab roundabout, check whether startNode is
+    // still an internal cycle node. If yes, accumulate regardless of item type.
+    // If no, we've exited — flush before normal processing.
```
truckermudgeon, line 123: > ```
> // If no, we've exited — flush before normal processing.
> ```

The telltale em-dash 🙃 

Last night I decided to give Claude Code a try.

So far: not bad. Seems to work well when giving it well-defined tasks (e.g., I knew I wanted to use an accumulation buffer). Just had to do a few minor course-corrections. For posterity, my prompts:

> i'm adding the ability to combine separate steps into one step that represents a maneuver within a roundabout. i think that means that i need to, in RouteStepBuilder::add, track when i've entered a roundabout, accumulate step data, then detect when i exit the roundabout so i can flush what i've accumulated into some composite step.

> does roundaboutWip.lastEndNode need to be a Node? can it just be a NodeUid, like entryNodeUid is?

> i forgot to mention one key thing: a composite roundabout maneuver involves traveling through non-roundabout prefabs and/or roads!


> what does it mean in flushRoundaboutWip if exit is null? when would that happen?

> well... what if a route starts at a roundabout entrance, but ends somewhere in the middle of the roundabout?

> maybe we can still assert _something_ in flushRoundaboutWip... if we've accumulated into a roundabout, then we know what inner nodes are in the roundabout.
> if we're flushing and we're not at the exit, can we at least assert that we're in an inner node?


> for multi-prefab roundabouts, can we break them up into two steps:
> - the first step is the entry. the direction/branch-type should still be the roundabout-specific one we have now.
> - the second step is the exit. for now, let's use a branch-type of 'merge'.

> one more change: the arrow points for the entry step should just be that of the entry prefab curve, and not that of the entire geometry

PR #132, opened 2026-05-01
generator: add data-subset command; io: read from FileSource/zip

truckermudgeon:
There's a few tests that are disabled in the codebase, e.g., this

https://github.com/truckermudgeon/maps/blob/e971a8b24c756983c8f0a4c309b0501dc726d367/packages/apis/navigation/domain/actor/tests/detect-route-events.test.ts#L204-L206

and this

https://github.com/truckermudgeon/maps/blob/e971a8b24c756983c8f0a4c309b0501dc726d367/packages/apis/navigation/domain/actor/tests/generate-routes.test.ts#L223-L225

This PR paves the way to address that, by:
- adding a `generator data-subset` command that reads parser-generated JSON, applies focus filtering, and writes that filtered data out somewhere
- updating `io` to support reading `parser`,`generator` data from zip files

(Note: I'm trialing Claude Code, and figured these tasks were good ones to throw at it.)

### DIFF
--- a/packages/clis/generator/commands/data-subset.ts
+++ b/packages/clis/generator/commands/data-subset.ts
@@ -1,0 +1,142 @@
+import type { MapDataKeys } from '@truckermudgeon/io';
+import { readMapData, writeArrayFile } from '@truckermudgeon/io';
+import fs from 'fs';
+import path from 'path';
+import type { Argv, BuilderArguments } from 'yargs';
+import { logger } from '../logger';
+import { parseFocusOptions } from './focus-options';
+import { maybeEnsureOutputDir, untildify } from './path-helpers';
+
+export const command = 'data-subset';
+export const describe =
+  'Reads parser-generated JSON files, applies focus filtering, and writes the filtered subset back out as JSON files.';
+
+const allMapDataKeys = [
+  // MapData
+  'nodes',
+  'elevation',
+  'roads',
+  'ferries',
+  'prefabs',
+  'companies',
+  'models',
+  'mapAreas',
+  'pois',
+  'dividers',
+  'trajectories',
+  'triggers',
+  'signs',
+  'cutscenes',
+  'cities',
+  // DefData
+  'countries',
+  'companyDefs',
+  'roadLooks',
+  'prefabDescriptions',
+  'modelDescriptions',
+  'signDescriptions',
+  'achievements',
+  'routes',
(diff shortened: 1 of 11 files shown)


packages/libs/io/file-source.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+}
+
+export function fromZip(zipPath: string): FileSource {
+  const zip = new AdmZip(zipPath);
+  // Index by basename so both "usa-roads.json" and "subset/usa-roads.json"
+  // entries resolve from the same lookup.
+  const entriesByName = new Map<string, AdmZip.IZipEntry>();
+  for (const entry of zip.getEntries()) {
+    if (entry.isDirectory) continue;
```
truckermudgeon, line 33: Note to self: update prettier config so that statements like these are re-written as

```ts
if (entry.isDirectory) {
  continue;
}
```

packages/clis/generator/commands/data-subset.ts
```diff
@@ -0,0 +1,142 @@
...
+import type { Argv, BuilderArguments } from 'yargs';
+import { logger } from '../logger';
+import { parseFocusOptions } from './focus-options';
+import { maybeEnsureOutputDir, untildify } from './path-helpers';
+
+export const command = 'data-subset';
+export const describe =
+  'Reads parser-generated JSON files, applies focus filtering, and writes the filtered subset back out as JSON files.';
+
+const allMapDataKeys = [
```
truckermudgeon, line 14: I'm surprised there already isn't a const for this 🤔

packages/clis/generator/commands/data-subset.ts
```diff
@@ -0,0 +1,142 @@
...
+      type: 'string',
+      conflicts: 'focusGameCoords',
+    })
+    .option('focusGameCoords', {
+      alias: 'c',
+      describe:
+        'Game coords to focus on, in "x,y" format (e.g., "-35578,20264"). Output files will only contain objects within `focusRadius`.',
+      type: 'string',
+      conflicts: 'focusCity',
+    })
```
truckermudgeon, line 66: These two options should be mutually exclusive

truckermudgeon: 🤦 I forgot: yargs' `conflicts` options already make them mutually exclusive.

PR #135, opened 2026-05-03
navigator: fix mobx strict-mode warnings

truckermudgeon:
During some manual testing, I noticed some browser console errors:

```
[MobX] Since strict-mode is enabled, changing (observed) observable values without using an 
action is not allowed. Tried to modify: AppStoreImpl@1.map
[mobx] Observable 'UiEnvironmentStoreImpl@3.handleResize' being read outside a reactive context. 
[mobx] Observable 'AppStoreImpl@1.map' being read outside a reactive context.
```

This PR fixes those. The general theme is: in ctors, read/write observables _before_ calling `makeAutoObservable`.

### DIFF
--- a/packages/apps/navigator/src/controllers/controls.ts
+++ b/packages/apps/navigator/src/controllers/controls.ts
@@ -10,13 +10,15 @@ import type {
 
 export class ControlsStoreImpl implements ControlsStore {
   bearing = 0;
-  units: 'imperial' | 'metric';
   limit = 0;
   speed = 0;
 
   constructor(private readonly appStore: AppStore) {
-    this.units = appStore.map === 'usa' ? 'imperial' : 'metric';
     makeAutoObservable(this);
+  }
+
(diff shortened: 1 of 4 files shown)


packages/apps/navigator/src/controllers/controls.ts
```diff
@@ -10,15 +10,17 @@ import type {
...
 
   constructor(private readonly appStore: AppStore) {
-    this.units = appStore.map === 'usa' ? 'imperial' : 'metric';
     makeAutoObservable(this);
   }
 
+  get units(): 'imperial' | 'metric' {
+    return this.appStore.map === 'usa' ? 'imperial' : 'metric';
+  }
+
```
truckermudgeon, line 23: Making this a readonly getter fell out of addressing the mobx warning... which then led to making the showXXXFab fields readonly getters, too. Yay more-honest, more-constrained interfaces 🎉

PR #136, opened 2026-05-04
  navigation,navigator: detect and auto-recover from stale telemetry bindings                                             

truckermudgeon:
This PR tries to improve the user experience around stale viewer-and-telemetry bindings.

When users start the Navigator webapp, they'll see:

image

<br/>

If, after 10 seconds, no telemetry data is detected (maybe because of some weird websocket/network blip, or because the webapp insists it's paired to a telemetry client that no longer exists), then they'll see:

image

<br/>

and can choose to try reconnecting, or try re-pairing to a client.

And, if they're in the middle of a trucking session and telemetry is lost (e.g., because the client crashed), then they'll see:

image


truckermudgeon (commented):
Revising comments and tests-as-documentation took longer than I expected / wanted to, so a few commits at the end of this PR were "manual". An interesting lesson 🤔

PR #142, opened 2026-05-14
navigation: reduce unnecessary reroutes

truckermudgeon:
One thing I've noticed in the Navigator metrics is period of time where telemetry pushes take longer than expected:

image

This happens because the backend ends up in a loop where it reroutes to _the same_ route over and over again, which can be especially expensive if the route is a long, cross-country route.

This PR puts a bandaid on the problem by:
- tagging `Route`s with information about the route's starting node + direction, then
- skipping recalculation of a re-route, if the truck is at the same starting node + facing the same starting direction used to calculate the active route

I say it's a bandaid because I'm pretty sure the real error is somewhere in the `isTruckOnRoute` function... I will look into that more, if this constant-rerouting problem persists.

### DIFF
--- a/packages/apis/navigation/domain/actor/detect-route-events.ts
+++ b/packages/apis/navigation/domain/actor/detect-route-events.ts
@@ -21,7 +21,7 @@ import type { GameContext } from '../game-context';
 import type { GraphAndMapData } from '../lookup-data';
 import type { TelemetryEventEmitter } from '../session-actor';
 import type { RouteWithLookup, RoutingService } from './generate-routes';
-import { reroute } from './generate-routes';
+import { calculateFromNodeAndDirection, reroute } from './generate-routes';
 import { scoreLine } from './score-line';
 
 export type RouteEventEmitter = EventEmitter<{
@@ -174,7 +174,6 @@ interface IsTruckOnRouteContext {
 
 function isTruckOnRoute(
   truck: TruckSimTelemetry['truck'],
-  activeRoute: RouteWithLookup,
   context: IsTruckOnRouteContext,
 ): boolean {
   const { roadAndPrefabRTree, nodes, expectedItems } = context;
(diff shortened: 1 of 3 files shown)


packages/apis/navigation/domain/actor/tests/detect-route-events.test.ts
```diff
@@ -197,7 +197,7 @@ describe('arrayIndexToRouteIndex', () => {
 
 // super expensive to run; disable for now until
 // data can be constrained to minimal fixtures.
```
truckermudgeon, line 199: I should really address this TODO 😬

packages/apis/navigation/domain/actor/detect-route-events.ts
```diff
@@ -218,27 +217,6 @@ function isTruckOnRoute(
     return true;
   }
 
-  // HACK special-case: look to see if we've just rerouted.
-  // rerouting can produce routes starting at the road/prefab _following_ the
-  // truck's current location. if the truck is stationary in this case, then we
-  // end up in a loop where we're always recalculating the same route, over and
-  // over again.
-  if (activeRoute.lookup.nodeUidsFlat.length === expectedItems.length + 1) {
```
truckermudgeon, line 226: This situation described in this comment, which involves recalculating the same route over and over again, should be handled by the new logic.

packages/apis/navigation/domain/actor/tests/detect-route-events.test.ts
```diff
@@ -197,7 +197,7 @@ describe('arrayIndexToRouteIndex', () => {
...
-describe.skip('detectRouteEvents bugs', () => {
+describe('detectRouteEvents bugs', () => {
```
truckermudgeon:
```suggestion
describe.skip('detectRouteEvents bugs', () => {
```

PR #143, opened 2026-05-15
Update extra-labels for Illinois

nautofon:
This proposed change addresses two issues visible in Illinois (for targets `il_is180`, `il_is74`, `il_mtvernon`).

I think cases like these ones will probably occur again elsewhere in future. So solving them by code changes rather than metadata changes seems sensible.


truckermudgeon (approved):
Thank you for the PR 🙂 

LGTM

PR #144, opened 2026-05-16
demo: calculate map extents at runtime

truckermudgeon:
This PR addresses a long-standing TODO in the `demo` webapp:

```ts
// TODO read these values from the .pmtiles files at runtime.
const extents = {
  ats: [
    [-124.477162, 25.767968].map(n => Math.floor(n)),
    [-88.928336, 49.122384].map(n => Math.ceil(n)),
  ].flat() as Extent,
  ets2: [
    [-10.025698, 34.897275].map(n => Math.floor(n)),
    [33.284941, 71.573102].map(n => Math.ceil(n)),
  ].flat() as Extent,
};
```

that has finally annoyed me enough to address (I was trying to debug some other issue related to Illinois, and wasn't able to get the game coordinates at a certain point, because these hard-coded extents were out of date)

### DIFF
--- a/packages/apps/demo/src/ContextMenu.tsx
+++ b/packages/apps/demo/src/ContextMenu.tsx
@@ -25,6 +25,7 @@ import {
   fromWgs84ToAtsCoords,
   fromWgs84ToEts2Coords,
 } from '@truckermudgeon/map/projections';
+import { getPmTilesBounds } from '@truckermudgeon/ui';
 import { distance } from '@turf/distance';
 import { featureCollection, point } from '@turf/helpers';
 import type { GeoJSON } from 'geojson';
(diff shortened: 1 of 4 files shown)


packages/libs/ui/pmtiles.ts
```diff
@@ -12,3 +13,24 @@ export function addPmTilesProtocol() {
   maplibregl.addProtocol('pmtiles', protocol.tile);
   pmTilesProtocolAdded = true;
 }
+
+const boundsCache = new Map<string, Promise<Extent>>();
```
truckermudgeon, line 17: Yes, yes... an unbounded cache 😐.

Should be fine. There are really on ever two URLs used by any given package in this repo.

packages/apps/demo/src/ContextMenu.tsx
```diff
@@ -50,20 +52,17 @@ interface ClickContext {
...
+  tileRootUrl: string;
+}
 
-export const ContextMenu = () => {
+export const ContextMenu = memo((props: ContextMenuProps) => {
+  const { tileRootUrl } = props;
   const map = assertExists(useMap().current);
+  const [extents, setExtents] = useState<{ ats: Extent; ets2: Extent }>({
+    ats: [0, 0, 0, 0],
+    ets2: [0, 0, 0, 0],
```
truckermudgeon, line 64: I guess this is technically a bug: a user could summon the context menu _before_ the `useEffect` to get the real extents completes... and then UI based on calculations would be off (e.g., the UI will say "click in the game area", even though they _did_ click in the game area).

But the chances of that happening in practice are slim to none.

PR #145, opened 2026-05-17
navigation: fix some memory leaks

truckermudgeon:
Looks like I've got some memory leaking:

image

This PR is an attempt to address three possible leaks around the `SessionActor`, which registers ~10 event listeners (and after auditing, looks like it doesn't always clean them up properly).

### DIFF
--- a/packages/apis/navigation/domain/session-actor.ts
+++ b/packages/apis/navigation/domain/session-actor.ts
@@ -186,6 +186,7 @@ export class SessionActorImpl implements SessionActor {
   dispose() {
     this.telemetryEventEmitter.off('telemetry', this.onTelemetryHandler);
     this.jobEventEmitter.removeAllListeners();
+    this.mapEventEmitter.removeAllListeners();
     this.routeEventEmitter.removeAllListeners();
     this.trailerEventEmitter.removeAllListeners();
     this.themeModeEventEmitter.removeAllListeners();
--- a/packages/apis/navigation/infra/actors/registry.ts
+++ b/packages/apis/navigation/infra/actors/registry.ts
@@ -20,6 +20,7 @@ import type { MetricsService } from '../metrics/service';
 interface Entry {
   actor: SessionActor;
   lastTouched: number;
+  dispose: () => void;
 }
 
 export interface ReadonlySessionActorRegistry {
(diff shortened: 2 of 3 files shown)


packages/apis/navigation/infra/actors/subscribe-session.ts
```diff
@@ -125,48 +128,58 @@ export function subscribeSession(
     actor.trailerEventEmitter.on('update', onTrailerUpdate);
     actor.themeModeEventEmitter.on('update', onThemeModeUpdate);
 
-    console.log(
-      'actor subscriptions',
-      actor.routeEventEmitter.listeners.length,
```
truckermudgeon, line 130: When digging through logs, I noticed this _always_ printed out `1`... because `EventEmitter.listeners` is a _function_ that takes _1_ argument. It is _not_ an array of listeners 🤦

packages/apis/navigation/infra/actors/registry.ts
```diff
@@ -79,7 +80,7 @@ export class SessionActorRegistryImpl implements SessionActorRegistry {
 
       let lastEmit: number | undefined;
 
-      this.opts.kv.onSet(event => {
+      const offSet = this.opts.kv.onSet(event => {
```
truckermudgeon, line 83: This was another 🤦 on my part: I was simply ignoring the cleanup function that gets returned.

packages/apis/navigation/domain/session-actor.ts
```diff
@@ -186,6 +186,7 @@ export class SessionActorImpl implements SessionActor {
   dispose() {
     this.telemetryEventEmitter.off('telemetry', this.onTelemetryHandler);
     this.jobEventEmitter.removeAllListeners();
+    this.mapEventEmitter.removeAllListeners();
```
truckermudgeon, line 189: Forgot to add this when I added ETS2 support 🫠

packages/apis/navigation/infra/actors/subscribe-session.ts
```diff
@@ -62,48 +61,51 @@ export function subscribeSession(
...
+    hasTelemetry = true;
+    wake?.();
+    wake = null;
   });
 
+  const onAbort = () => {
+    wake?.();
+    wake = null;
+  };
+  signal?.addEventListener('abort', onAbort, { once: true });
```
truckermudgeon, line 107: Lesson learned: never ignore `AbortSignal`

